PR #1, opened 2015-03-05
Some cleanup in PyQtAwesome

SylvainCorlay:
Separated font-awesome specific things from general font-based icons.

### DIFF
--- a/QtAwesome/iconic_font.py
+++ b/QtAwesome/iconic_font.py
@@ -1,6 +1,10 @@
+"""Classes handling iconic fonts"""
+
 from __future__ import print_function
-from PyQt4.QtCore import Qt, QPoint, QRect, qRound
-from PyQt4.QtGui import QIcon, QIconEngine, QPainter, QPixmap
+from PyQt4.QtCore import Qt, QObject, QChar, QPoint, QRect, qRound, QByteArray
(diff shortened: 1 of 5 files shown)


example.py
```diff
@@ -26,37 +26,30 @@ class AwesomeExample(QtGui.QWidget):
```
goanpeca: A QDialog will center in the screen so maybe it is better to change to 

`class AwesomeExample(QtGui.QDialog)`

QtAwesome/iconic_font.py
```diff
@@ -85,42 +56,74 @@ def pixmap(self, size, mode, state):
...
+        names to char numbers"""
+        super(IconicFont, self).__init__()
+        self.painter = CharIconPainter()
+        self.codes = codes
         self.painters = {}
-        _load()
+        self.ttf_name = ttf_filename
+        self._load_font()
+
+    def _load_font(self):
```
goanpeca: In the latest version the `_load_font` is no longer "private" (`load_font`). I would leave it private as this is something that would not be directly exposed to the user... so better to have `_load_font`

SylvainCorlay: I think that this should be solved in master.

PR #2, opened 2015-03-09
Allow stacking font icons

SylvainCorlay:
Using the `icon_stack` function

### DIFF
--- a/example.py
+++ b/example.py
@@ -4,6 +4,7 @@ import QtAwesome as qta
 
 
 class CustomIconPainter:
+
     """A custom painter for example bellow"""
     def paint(self, awesome, painter, rectIn, mode, state, options):
         drawSize = QtCore.qRound(rectIn.height() * 0.5)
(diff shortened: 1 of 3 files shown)


example.py
```diff
@@ -44,16 +45,24 @@ def __init__(self):
         
         # Get icons by character
         coffee_icon = qta.icon_by_char('fa', qta.charmap('fa.coffee'))
-        coffee_button = QtGui.QPushButton(coffee_icon, 'Black please!')
+        coffee_button = QtGui.QPushButton(coffee_icon, 'Access by character')
         
         # Render a label with this font
-        label = QtGui.QLabel(QtCore.QChar(0xf0f4))
+        label = QtGui.QLabel(QtCore.QChar(0xf19c) + ' ' + 'Label')
```
goanpeca: @SylvainCorlay, I know you changed this to `unichr`, but is there any real use case (for the main users) for directly using the unicode number of the icon, as opposed to the just the name?

PR #3, opened 2015-03-11
Add support for pulse and spin rotation of icons

goanpeca:


QtAwesome/iconic_font.py
```diff
@@ -10,12 +10,20 @@
 _default_options = {
     'color': QColor(50, 50, 50),
     'scale_factor': 0.9,
+    'rotation': False,      # Valid options are: 'spin', or 'pulse'
```
SylvainCorlay: You can probably avoid setting `rotation` or `parent` at all since you are using `dict.get`

QtAwesome/iconic_font.py
```diff
@@ -10,12 +10,20 @@
...
+    'parent': False,        # Ref to instance of widget holding the icon
+    'spin_interval': 10,    # miliseconds
+    'spin_step': 1,         # degrees
+    'pulse_interval': 300,  # miliseconds
+    'pulse_step': 45,       # degrees
```
SylvainCorlay: why not simply having a single dict attribute `animation`. 
```
'animation': {
    'type': 'pulse',  # or 'rotation'
    'step': 1,
    'interval': ...,   
}
```

QtAwesome/iconic_font.py
goanpeca: Sure, I thought it made the painter cleaner, but valid :)

QtAwesome/iconic_font.py
goanpeca: Yes I also considered this, but on stacked icons it would be a dict inside a dict inside a list... so it felt a bit obfuscated.

Still, pulse and spin should have default values of step and speed, so where should I store this then?

SylvainCorlay: Re-thinking about it, rather than having our generic painter implement the animation, the `animation` option could be an object implementing the animation. The painter would generically call
```python
animation = options.get('animation')
if animation is not None:
    animation.setup(self)
```
and this lets the door open to other things later without bloating the painter.

goanpeca: Yes I like it :), for shadow and for offset we could do the same,

goanpeca: I am now on windows and I am getting the "Font is empty" message... with the code unchanged

Ideas?  ?

goanpeca: @SylvainCorlay ok we have an issue.. on windows the icons look horrible!

SylvainCorlay: Well, styling properties belong more naturally to the painter than animations.
I guess that if someone wants to change styling behavior, the right thing to do is to use a custom painter. Hence I would not move the offset or rotation. I have some changes regarding offset, opacity, I just need to push.

SylvainCorlay: Regarding the aspect, I think that it is because ttf embed an already rendered bitmap for certain sizes. If you want to have a nice rendering for other sizes, we should use crispEdges styling. I did not do it because I could not see the problem occur.

goanpeca: I will update the PR with the new changes... but what should be donde about this then?

image

QtAwesome/iconic_font.py
```diff
@@ -30,22 +89,28 @@ def _paint_icon(self, awesome, painter, rect, mode, state, options):
...
             char = options.get('active', char)
-        elif(mode == QIcon.Selected):
+        elif mode == QIcon.Selected:
             color = options.get('color_selected', color)
             char = options.get('selected', char)
 
         painter.setPen(QColor(color))
 
-        drawSize = qRound(rect.height() * options['scale_factor'])
+        rec_size = rect.height()
```
SylvainCorlay: Since you reverted this part, you don't need a local variable anymore.

QtAwesome/iconic_font.py
```diff
@@ -1,22 +1,81 @@
...
 
+
 _default_options = {
     'color': QColor(50, 50, 50),
     'scale_factor': 0.9,
 }
 
 
-class CharIconPainter:
+class CharIconAnimator:
```
SylvainCorlay: CharIconPulse / CharIconRotation... And no `if` check for 'pulse' or 'rotation?'

QtAwesome/iconic_font.py
goanpeca: Ok

QtAwesome/iconic_font.py
goanpeca: *  Change name to CharIconRotation :+1: 
* As I said before spin and pulse are the names used in font awesome, that is why I left the same names, so pulse and spin refer to font awesome `fa-spin fa-pulse`

Some icons allow for a pulsing motion, and any icon could be used for spinning (plus I can add direction: clockwise, counterclockwise) some have a sniping motion. By default we should handle these `types` without needing to define the step and the intervals manually. Maybe in Gitter you can explain me :)

goanpeca: Did you suggest to create a separate class for Pulse and a separate class for a generic rotation?

SylvainCorlay: Yes. I think that having a check for a `type` string is not very object-oriented :)

goanpeca: Maybe, but so far I do not see any gain... right now I am storing the info of the animations inside the `CharIconPainter` in `self.animator = CharIconAnimator()`, using 2 classes for that seems overkill..., I would still need to know which class to initialize, so maybe I can change and if for a dictionary, but there will be some choosing...

SylvainCorlay: Yes, we can certainly do that, Do you want me to open a PR to your PR tonight to show what I mean?

SylvainCorlay: btw, this is *really* cool!

goanpeca: Well if you have a clear idea of what to do, then maybe you can merge and work over what I have done so far :) ?

goanpeca: Still, unfortunately, the spinning stuff looks bad... stupid icons are not centered...

SylvainCorlay: yeah, I should do that. I have a meeting now, and will merge it when I am back.

PR #5, opened 2015-03-12
Support offset and opacity

SylvainCorlay:
...

### DIFF
--- a/QtAwesome/iconic_font.py
+++ b/QtAwesome/iconic_font.py
@@ -43,8 +43,7 @@ class CharIconPainter:
 
         painter.setPen(QColor(color))
 
-        rec_size = rect.height()
-        draw_size = qRound(rec_size * options['scale_factor'])
+        draw_size = qRound(rect.height() * options['scale_factor'])
         prefix = options['prefix']
 
         # Animation setup hook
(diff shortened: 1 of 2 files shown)


goanpeca:
Hi @SylvainCorlay, cool :) I would say that it would be very helpful to have a shadow keyword that would under the curtains do an stack offset with some default parameters (offset plus color). It would be a convenience thing that would make creating a shadowed icon super easy.

goanpeca:
So instead of

```python
# Stack and offset icons
saveall = qta.icon_stack(['fa.save', 'fa.save'],
                                      options=[{'scale_factor': 0.8, 
                                             'offset': (0.2, 0.2),
                                             'color': 'gray'},
                                            {'scale_factor': 0.8}])

```

```python
# Stack and offset icons
        saveall = qta.icon('fa.save', shadow=True)

```

goanpeca:
oops sorry, hit close by accident!!

SylvainCorlay:
OK, this is be a good idea to add higher-level APIs for specific effects like shadow.

goanpeca:
Yep, high level api sound good. In this sense it would make sense to be able to define global defaults for this high level api, so the user can set the shadow defaults once when loading pyqtawesome

SylvainCorlay:
Rebased

PR #6, opened 2015-03-12
Animation objects

SylvainCorlay:
This is separating animations in their own files, and slightly changing the API for setting animations. 

Besides default values, `Spin` and `Pulse` seem to be doing the same thing, so I will refactor in next commit unless I missed something. 
I don't like that animations require the parent widget when creating them, I think that this should dynamically grab it when painting. Looking at it.

### DIFF
--- a/QtAwesome/animation.py
+++ b/QtAwesome/animation.py
@@ -1,0 +1,41 @@
+from PyQt4.QtCore import QTimer
+
+
+class Spin:
+
+    def __init__(self, parent_widget, interval=10, step=1):
+        self.parent_widget = parent_widget
+        self.interval, self.step = interval, step
+        self.info = {}
+
+    def _update(self, parent_widget):
+        if self.parent_widget in self.info:
+            timer, angle, step = self.info[self.parent_widget]
+
+            if angle >= 360:
+                angle = 0
+
+            angle += step
+            self.info[parent_widget] = timer, angle, step
+            parent_widget.update()
+
+    def setup(self, icon_painter, painter, rect):
+
+        if self.parent_widget not in self.info:
+            timer = QTimer()
+            timer.timeout.connect(lambda: self._update(self.parent_widget))
+            self.info[self.parent_widget] = [timer, 0, self.step]
+            timer.start(self.interval)
+        else:
+            timer, angle, self.step = self.info[self.parent_widget]
+            x_center = rect.width() * 0.5
+            y_center = rect.height() * 0.5
+            painter.translate(x_center, y_center)
+            painter.rotate(angle)
+            painter.translate(-x_center, -y_center)
+
+
+class Pulse(Spin):
+
+    def __init__(self, parent_widget):
+        Spin.__init__(self, parent_widget, interval=300, step=45)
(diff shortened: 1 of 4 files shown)


goanpeca:
Yes they are the same... so if you want to keep it OO then maybe something like

```python
class BaseRotation:

    def __init__(self, parent_widget, interval=None, step=None):
        self.parent_widget = parent_widget
        self.interval, self.step = interval, step
        self.info = {}

    def _update(self, parent_widget):
        if self.parent_widget in self.info:
            timer, angle, step = self.info[self.parent_widget]

            if angle >= 360:
                angle = 0

            angle += step
            self.info[parent_widget] = timer, angle, step
            parent_widget.update()

    def setup(self, icon_painter, painter, rect):

        if self.parent_widget not in self.info:
            timer = QTimer()
            timer.timeout.connect(lambda: self._update(self.parent_widget))
            self.info[self.parent_widget] = [timer, 0, self.step]
            timer.start(self.interval)
        else:
            timer, angle, self.step = self.info[self.parent_widget]
            x_center = rect.width() * 0.5
            y_center = rect.height() * 0.5
            painter.translate(x_center, y_center)
            painter.rotate(angle)
            painter.translate(-x_center, -y_center)


class Spin(BaseRotation):
    def __init__(self, parent_widget, interval=10, step=1):
        BaseRotation.__init__(self, parent_widget)
        self.interval, self.step = interval, step

class Pulse(BaseRotation):
    def __init__(self, parent_widget, interval=300, step=45):
        BaseRotation.__init__(self, parent_widget)
        self.interval, self.step = interval, step
```

goanpeca:
I dont like hacing to give the parent also... but at this moment I have no clue how to do that dynamically

SylvainCorlay:
or `Rotation.__init__(self, parent_widget, interval, step)`

goanpeca:
I am not entirely in love with....

```python

         pulse_icon = qta.icon('fa.spinner', color='green',
                              animation=qta.Pulse(pulse_button))
```

somehow  (if we manage to get rid of passing the parent!)

```python
         pulse_icon = qta.icon('fa.spinner', color='green',
                              animation='pulse')
```

feels better

SylvainCorlay:
I would have done `animation=qta.Pulse()` if I did not need the parent widget. You approach requires the icon function to infer the animation object from a string, that is to know of any possible animation.

QtAwesome/animation.py
```diff
@@ -0,0 +1,41 @@
...
+            x_center = rect.width() * 0.5
+            y_center = rect.height() * 0.5
+            painter.translate(x_center, y_center)
+            painter.rotate(angle)
+            painter.translate(-x_center, -y_center)
+
+
+class Pulse(Spin):
+
+    def __init__(self, parent_widget, interval=300, step=45):
```
goanpeca: a Pulse (at least in font awesome) only makes sense for elements with 8 fold symmetry (hence the 45 angle... so maybe the call should be: 
```python
def __init__(self, parent_widget):
    interval=300, step=45
    Spin.__init__(self, parent_widget, interval, step)
```

SylvainCorlay: Good point.

goanpeca:
Then just for the sake of usage we should encourage

```python

import QtAwesome as qta
from QtAwesome import Pulse, Spin

pulse_icon = qta.icon('fa.spinner', color='green', animation=Pulse(pulse_button))
```

PR #11, opened 2015-03-18
Fix Travis url in README

ccordoba12:
I don't know why you selected camel case for the project name on PyPi, etc, (I really don't like it) but @SylvainCorlay agreed with a lower case name for the repo ;-)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,7 +13,7 @@
 .. image:: https://pypip.in/license/QtAwesome/badge.svg
 
 .. image:: https://travis-ci.org/spyder-ide/QtAwesome.svg?branch=master
-   :target: https://travis-ci.org/spyder-ide/QtAwesome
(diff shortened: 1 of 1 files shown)


goanpeca:
@ccordoba12, it is just the name, we think with @SylvainCorlay that it looked better that way. What I do agree on changing is that the install example should be

`pip install qtawesome` as it is more clear, since pip ignores casing AFAIK

SylvainCorlay:
Well this is just a URL change. I am always hesitant between CamelCase and PEP8 when interacting with a library like Qt which uses CamelCase. (For example, when inheriting etc.)

ccordoba12:
I also fixed the image link, but I commited that change directly to master this time. I hope it's ok for such a simple fix :-)

SylvainCorlay:
Thanks! I still don't know why the tests are not getting triggered.

PR #13, opened 2015-03-19
fix for python 2.6, dictionary comprehensions are only available in >2.7

sebastianelsner:


qtawesome/iconic_font.py
```diff
@@ -126,14 +126,20 @@ def load_font(self, prefix, ttf_filename, charmap_filename, directory=None):
         directory: str or None, optional
             directory for font and charmap files
         """
+
+        def hook(o):
+            r={}
+            for k,v in o.iteritems():
+                r[k]=unichr(int(v, 16))
```
goanpeca: can you use pep8 ? space before and after `=`

qtawesome/iconic_font.py
```diff
@@ -126,14 +126,20 @@ def load_font(self, prefix, ttf_filename, charmap_filename, directory=None):
         directory: str or None, optional
             directory for font and charmap files
         """
+
+        def hook(o):
+            r={}
+            for k,v in o.iteritems():
```
goanpeca: Is this python3 compatible?

We are using [six](https://pythonhosted.org/six/#six.iteritems) so maybe you can adjust so that it is Py2/3 compatible as well?

SylvainCorlay:
Thanks for the fix. Besides the formatting issue, I am fine with this. Besides, `iteritems` does not exist in Python 3 anymore, but we already import `six`, which defines a unique function for Python 2 and 3:
https://pythonhosted.org/six/

sebastianelsner:
Thanks for the comments, pep8 was somehow not working in my PyDev (but now I get additional warnings at other places, too)

PR #14, opened 2015-03-19
pep8

sebastianelsner:
fix all pep8 warnings

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,7 +1,10 @@
 import sys
+
 from qtpy import QtGui, QtCore
-import qtawesome as qta
 from six import unichr
+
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
I am not sure about this one. 

Imports with parentheses like
```python
from qtpy.QtGui import (QIcon, QColor, QIconEngine, QPainter, QPixmap,		
                        QFontDatabase, QFont)
```
are ok in pep8 and it is what we do in Spyder. 

Since this includes your other merged PR, this would need to be rebased on the latest master.

sebastianelsner:
ah sorry, I will fix this

PR #15, opened 2015-03-19
Fix qt5 build issues

SylvainCorlay:
@goanpeca

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,5 +1,5 @@
 import sys
-from qtpy import QtGui, QtCore
+from qtpy import QtGui, QtCore, QtWidgets
 import qtawesome as qta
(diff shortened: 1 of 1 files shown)


goanpeca:
And maybe the push buttons are also imported from QtWidgets...

SylvainCorlay:
I am still waiting for the build... It is weird that it takes so much time.

SylvainCorlay:
We are almost there.

goanpeca:
@Sylvain, I think that in order to better get used to these changes.. and to change them more easily in the future (spyder included) we should import the needed items on top (like on spyder)

```from PyQt.QtWidgets import QPushButton``` 

etc

PR #32, opened 2015-11-30
Adding release instruction and setting current version to dev

SylvainCorlay:
@ccordoba12 streamlining release process for qtawesome.

I also added tags for the past versions, from `0.1.5`. I don't think that earlier versions have much interest.

### DIFF
--- a/qtawesome/_version.py
+++ b/qtawesome/_version.py
@@ -1,2 +1,2 @@
-version_info = (0, 1, 11)
+version_info = (0, 2, 0, 'dev0')
 __version__ = '.'.join(map(str, version_info))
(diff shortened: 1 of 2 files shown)


qtawesome/_version.py
```diff
@@ -1,2 +1,2 @@
-version_info = (0, 1, 11)
+version_info = (0, 2, 0, 'dev')
```
ccordoba12: Should we use PEP 440 conventions for versions? I mean, in that case this should read `dev0` instead of `dev` (I think :-)

SylvainCorlay: hum, google says you are right :)

SylvainCorlay: fixed.

PR #37, opened 2015-12-05
Use high DPI pixmaps

SylvainCorlay:
Using `app.setAttribute(Qt.AA_UseHighDpiPixmaps)`, which only exists with PyQt5.


ccordoba12:
Seems good to me, merging :-)

PR #43, opened 2016-03-10
PR: Feature/add state handling

goanpeca:
This PR completes the expected behavior of icons, which include the possibility of a [State (On, Off)](http://doc.qt.io/qt-4.8/qicon.html#details)

Added new example to illustrate usage.


goanpeca:
@SylvainCorlay we kind of need this urgently, any chance you get to review today?

SylvainCorlay:
Hi Gonzalo,

I can look at it tonight but if you really need it *merged* now to iterate,
you should gi ahead and hit the green button.

Cheers,

S.
On Mar 10, 2016 9:49 AM, "Gonzalo Peña-Castellanos" <
notifications@github.com> wrote:

@SylvainCorlay <https://github.com/SylvainCorlay> we kind of need this
urgently, any chance you get to review today?

—
Reply to this email directly or view it on GitHub
<https://github.com/spyder-ide/qtawesome/pull/43#issuecomment-194883095>.

goanpeca:
@SylvainCorlay what happen with I can take a look at it later tonight :laughing:?, its ok, been there as well, right @nodd?

Anyway, we are using our own icons (svg) directly, so this change is no longer critical... still I would appreciate if you took some time to review it as this covers an important use case.

Thanks :smile_cat:

SylvainCorlay:
Hi I am travelling this week so I won't be able to look at it right away. If you are confident, you can go ahead and merge. Regarding the svg icons, is it something you think should/could be integrated in the spyder font?

goanpeca:
>Regarding the svg icons, is it something you think should/could be integrated in the spyder font?

Nope this is very specific to the application we are developing and not related to Spyder.

qtawesome/iconic_font.py
```diff
@@ -13,12 +13,15 @@
...
 
+# Standard library imports
 from __future__ import print_function
-from qtpy.QtCore import Qt, QObject, QPoint, QRect, qRound, QByteArray
-from qtpy.QtGui import (QIcon, QColor, QIconEngine, QPainter, QPixmap,
-                        QFontDatabase, QFont)
 import json
 import os
+
+# Third party importd
```
Nodd: Typo in "importd"

qtawesome/iconic_font.py
```diff
@@ -56,17 +67,35 @@ def paint(self, iconic, painter, rect, mode, state, options):
...
+                char = options['off']
+            elif mode == QIcon.Disabled:
+                color = options['color_off_disabled']
+                char = options['off_disabled']
+            elif mode == QIcon.Active:
+                color = options['color_off_active']
+                char = options['off_active']
+            elif mode == QIcon.Selected:
+                color = options['color_off_selected']
+                char = options['off_selected']
```
Nodd: How about using dicts here instead of those elif ? This part could be written as `color, char = color_options[state][mode]`, and you'd have to fill the dict somewhere else. It would simplify the logic of the code inside the function.

qtawesome/iconic_font.py
```diff
@@ -206,17 +236,70 @@ def _parse_options(self, specific_options, general_options, name):
...
+                      'color_on': color_on,
+                      'color_on_active': color_on_active,
+                      'color_on_selected': color_on_selected,
+                      'color_on_disabled': color_on_disabled,
+                      'color_off': color_off,
+                      'color_off_active': color_off_active,
+                      'color_off_selected': color_off_selected,
+                      'color_off_disabled': color_off_disabled,
+                      }
+        options.update(color_dict)
```
Nodd: Why the `color_dict` here ? The lines above could just be like `options.setdefault('color_active', color_on)`. You'd have half the number of lines and less possibilities of confusion.

goanpeca: > why the color_dict here ?

I just changed the code where it was, @SylvainCorlay can answer why...

How can I get the lines reduced?, I still need to define the items...

qtawesome/iconic_font.py
goanpeca: I guess... want to do it :-) ?

qtawesome/iconic_font.py
Nodd: `setdefault` defines the item if it doesn't exist, so instead of getting the options, building a dict and updating the options, you directly set them if needed. (Or I didn't understand what the code does)

goanpeca: Yes you are right, setdefault can do the job :+1:

qtawesome/iconic_font.py
Nodd: I wouldn't interfere in your PR ! :tongue:

ccordoba12:
I addressed all @Nodd's comments, so I'm merging :-)

PR #50, opened 2016-09-25
Conda installation instructions

SylvainCorlay:


ccordoba12:
`qtawesome` is part of the default conda channels now (because of Spyder 3.0) so IMHO there's no need to direct people to conda-forge ;-)

PR #52, opened 2016-10-25
Enable use of Qt.py and use camelCase

dgovil:
Code now uses Qt.py as the default shim for abstracting Python Qt bindings.
This means that it now should support PySide2 in applications like Maya

Also I've replaced all (or most) uses of snake_case objects with camelCase.
This is so that it is more uniform within a Qt code base that favors camelCase


dgovil:
Sorry was trying to merge this into my own fork and github decided to push it to the master. Ignore this

SylvainCorlay:
out of curiosity,  why not using qtpy?

dgovil:
Mostly to support our studios choice of qt lib which is Qt.py since it supports pyside2 already which is important to us. 

Nothing against qtpy, in fact if there isnt a ticket for it, I'll probably make a branch of qtpy and add Pyside2 support.

goanpeca:
Hi, there is a plan for supporting pyside2 but las I heard it was not ready yet. If you could start working on providing pyside2 as a PR it would be awesome.

Thanks

ccordoba12:
@dgovil, the right way to go is for you to add support for PySide2 in `qtpy`. We're not going to use another shim for QtAwesome because it's a core dependency of Spyder, and Spyder already uses `qtpy` :-)

dgovil:
@goanpeca pyside2 is somewhat ready in that it is now shipping with commercial software, ie Maya 2017. I'll try and make a branch and pull request when I get some time though. 

@ccordoba12 yes I don't expect or desire qt awesome to use another shim. But our studio and a few others are using the other shim so I was just making a custom cut that uses it first and falls back to qtpy if it doesn't exist. 

This pull request was just a mistake because I was trying to send it to my own master fork and accidentally selected the original spyder fork.

PR #67, opened 2017-01-19
PR: Add a way to not verify hash of vendorized fonts

### LINKED ISSUE #64: Integration with system fonts
With the new release of QtAwesome v4.x, patching the code to use the system fonts produces a `FontError`, most likely due to a mismatch between the mdf5 sums of the vendored fonts compared to the system ones. Please remember that vendoring is *strongly* discouraged by Linux distributions, hence the need for an integration path.

I would like to discuss with you guys what the best course of actions would be to have `QtAwesome` work with system fonts without complicating the package maintenance burden. Shall, I just make a patch to disable the sum checking part of the code?

ccordoba12:
Fixes #64.

----

This is needed for Linux packagers that want to use system fonts.

### DIFF
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -26,6 +26,11 @@ from qtpy.QtGui import (QColor, QFont, QFontDatabase, QIcon, QIconEngine,
 from six import unichr
 
 
+# Linux packagers, please set this to True if you want to make qtawesome
+# use system fonts
+SYSTEM_FONTS = False
+
+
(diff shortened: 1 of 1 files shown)


ccordoba12:
@ghisvail, is this enough for you?

ghisvail:
Sounds good to me :+1:

ccordoba12:
Great! I'll release 0.4.2 in a couple of hours then :-)

ghisvail:
Awesome, thanks.

PR #68, opened 2017-01-22
PR: Include example script into tarball to run tests when the package is installed

ghisvail:
The script is needed for testing purposes on the packaging side.


ghisvail:
Would you consider releasing `0.4.3` with the above change? Without this script, the tests cannot be run during packaging which is quite a regression on our end.

Thanks,
Ghis

ccordoba12:
Ok, working on it :-)

PR #69, opened 2017-01-23
PR: Add a QApplication instance before calling load_font -> QFontDatabase

dalthviz:
This fixes the issue [#4014](https://github.com/spyder-ide/spyder/issues/4014) in Spyder. A segfault issue caused for the use of QFontDatabase without a QApplication instance when listing the `modules` with the `help()` command of the Python console.

Edit to include a reference to the issue in Spyder.

### DIFF
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -23,10 +23,8 @@ import hashlib
 from qtpy.QtCore import QObject, QPoint, QRect, qRound, Qt
 from qtpy.QtGui import (QColor, QFont, QFontDatabase, QIcon, QIconEngine,
                         QPainter, QPixmap)
+from qtpy import QtWidgets
 from six import unichr
-
-# Spyder imports
-from spyder.utils.qthelpers import qapplication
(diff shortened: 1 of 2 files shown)


qtawesome/iconic_font.py
```diff
@@ -174,6 +176,7 @@ def __init__(self, *args):
         self.painters = {}
         self.fontname = {}
         self.charmap = {}
+        qapplication()
```
goanpeca: @dalthviz you are making QTAwesome depend on Spyder, which is not what we want.

We should use QApplication

```python
from qtpy import QtWidgets

# ...

app = QApplication([])

```

ccordoba12:
@dalthviz, this breaks all our tests here. Please review.

dalthviz:
@ccordoba12 @goanpeca I think that the problem is not here, is more like that a third-party plugin (spyder-profiler) is doing a import of qtqwesome without/before a QApplication instance. So, I'm going to close this PR. Thanks for the comments :)

PR #70, opened 2017-01-25
PR: Prevent segfaults when importing QtAwesome out of a of QApplication

dalthviz:
Fixes spyder-ide/spyder#4014
Fixes spyder-ide/spyder#4049


qtawesome/iconic_font.py
```diff
@@ -247,17 +251,18 @@ def icon(self, *names, **kwargs):
...
+                                                          general_options,
+                                                          names[i]))
 
-        # Process high level API
-        api_options = parsed_options
+            # Process high level API
+            api_options = parsed_options
 
-        return self._icon_by_painter(self.painter, api_options)
+            return self._icon_by_painter(self.painter, api_options)
```
goanpeca: Maybe the else should return an empty QIcon plus some [warning](https://docs.python.org/2/library/warnings.html#warnings.warn)?

```
else:
   warning.warn('You need to have a running QApplication to use QtAwesome!')
   return QIcon()
```

goanpeca:
@dalthviz could we create a test that does the direct import (Like what @jitseniesen pointed out)

Just create a new tests folder inside the main module, ok?

qtawesome/tests/test_qtawesome.py
```diff
@@ -0,0 +1,14 @@
...
+Tests for QtAwesome
+"""
+# Standard library imports
+import subprocess
+
+# Test Library imports
+import pytest
+
+def test_segfault_import():
+    subprocess.call('python -c qtawesome')
```
ccordoba12: Please this to

    output_number = subprocess.call('python -c qtawesome')
    assert  output_number == 0

Remember that you always have to `assert` something on tests :-)

goanpeca: Should we try to import every file in the submodule also?

qtawesome/iconic_font.py
ccordoba12: I think adding the warning is fine ;-)

qtawesome/tests/test_qtawesome.py
```diff
@@ -0,0 +1,14 @@
+r"""
+Tests for QtAwesome
```
goanpeca: dot at the end of a docstring (it is a sentence)

ccordoba12: We don't have to be too strict about style on test files, but sure, why not :-)

qtawesome/tests/test_qtawesome.py
ccordoba12: The most important import's for QtAwesome are on its `__init__.py`, so I don't see the need for that.

ccordoba12:
@ghisvail, this is kind of important in case you want to pick it up for Debian.

ccordoba12:
@dalthviz, please also add the new tests folder to `MANIFEST.in`, like we do it in `qtpy`:

https://github.com/spyder-ide/qtpy/blob/master/MANIFEST.in#L5

That's necessary to add our tests to our released packages too, which was requested by @ghisvail for Debian.

goanpeca:
Then can we also update the CI code to run the test?

https://github.com/spyder-ide/qtawesome/blob/master/appveyor.yml#L60
`- "%CMD_IN_ENV% py.test qtawesome"`

https://github.com/spyder-ide/qtawesome/blob/master/circle.yml
```
    - export PATH="$HOME/miniconda/bin:$PATH" && source activate test && py.test qtawesome: # note the colon
        parallel: true
```

ccordoba12:
Ok, sure. All of that have to be part of this PR.

@dalthviz, sorry for putting you more work ;-)

ghisvail:
I can cherry-pick this, but have to wait for the last upload of qtawesome-0.4.3 to transition with the rest of the stack, i.e. qtpy-1.2.1 and spyder-3.1.2. Otherwise, none of them will make it for the freeze.

ccordoba12:
@dalthviz, I know why our tests are failing. I'll help you to fix them :-)

ccordoba12:
Ok, I think this one is ready! Thanks @dalthviz for your hard work to solve this nasty problem :-)

Issue #71, opened 2017-01-30
Enhancement: Add support for fa 4.7


amal-khailtash:
Please upgrade to fa 4.7 for extra icons...

SylvainCorlay:
Fixed in #72

PR #72, opened 2017-02-26
Updating FontAwesome to 4.7.

phargogh:
Greetings!  This PR merely updates FontAwesome to version 4.7, which would grant access to the latest icons.  This would also address Issue #71.

Happy to discuss, and thanks for all your work with `qtawesome`!


ccordoba12:
Thanks a lot for this!! To make our tests pass, you need to update FontAwesome's md5 hash here:

https://github.com/spyder-ide/qtawesome/blob/master/qtawesome/iconic_font.py#L230

phargogh:
Ah, of course!  I just pushed a new commit with the updated md5 digest of the included font.  Hopefully that'll cause the tests to pass 😄

PR #81, opened 2017-10-02
PR: Replace qRound (missing in PySide2) with round

gmarull:
PySide2 misses que qRound method. It can be simply replaced by the Python built-in round.

### DIFF
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -21,7 +21,7 @@ import hashlib
 import warnings
 
 # Third party imports
-from qtpy.QtCore import QObject, QPoint, QRect, qRound, Qt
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks for your contribution. Quick question: does `round` exist in Python 2 and 3?

gmarull:
Yes, it is available on both Python versions:

https://docs.python.org/2/library/functions.html#round
https://docs.python.org/3/library/functions.html#round

imatu5:
Hello, 

Is it possible to merge his PR as is it pretty useful :+1: 

Thanks,

ccordoba12:
Sure, thanks for remembering us about it.

PR #84, opened 2018-05-09
PR: Add compatibility with PySide2 without impacting PyQt5

stonebig:


ccordoba12:
Does this supersedes PR #81?

stonebig:
Yes. In the process of making PySide2 more usefull, and so getting more users fighting bugs:
- I converted guidata to 'QtPy' vendored,
- it required this patch.

Now: 
- I have a guidata working, (at 80%)
- but Spyder still doesn't start.

stonebig:
hoping this patch is more acceptable than previous one you rejected

ccordoba12:
Let's use #81, which is simpler.

PR #92, opened 2018-08-28
Add Material Design Icons support

doudz:


darkvertex:
First of all, awesome! :tada:

Which icon set are these? Google's or the other community project?

Could you please at least mention here where you sourced the icons from and what version they were, if they're versioned? Just for posterity. Thanks! :+1:

doudz:
This is the version 2.6.95 of MDI downloaded from https://materialdesignicons.com/
The TTF has been extracted from the zip archive
The json charmap has been made manually based on materialdesignicons.css and applying a bunch of find and replace operations ;)

doudz:
Just realize it comes from https://github.com/Templarian/MaterialDesign-Webfont

PR #101, opened 2018-10-23
PR: Generate PyPI long description from README.md

doudz:


setup.py
```diff
@@ -13,37 +13,15 @@
...
-.. code-block:: python
-
-    # Get icons by name.
-    fa_icon = qta.icon('fa5s.flag')
-    fa_button = QtGui.QPushButton(fa_icon, 'Font Awesome!')
-
-    asl_icon = qta.icon('ei.asl')
-    elusive_button = QtGui.QPushButton(asl_icon, 'Elusive Icons!')
-"""
+with open(path.join(here, 'README.md'), encoding='utf-8') as f:
```
ccordoba12: Please change `path.join` to `os.path.join` here. That's why our tests are failing.

doudz:
oops, fixed !

setup.py
doudz: fixed

setup.py
```diff
@@ -13,37 +13,15 @@
...
-.. code-block:: python
-
-    # Get icons by name.
-    fa_icon = qta.icon('fa5s.flag')
-    fa_button = QtGui.QPushButton(fa_icon, 'Font Awesome!')
-
-    asl_icon = qta.icon('ei.asl')
-    elusive_button = QtGui.QPushButton(asl_icon, 'Elusive Icons!')
-"""
+with open(os.path.join(here, 'README.md'), encoding='utf-8') as f:
```
ccordoba12: `encoding` is not a valid keyword for `open` in Python 2. I think you can use `io.open` instead.

PR #105, opened 2018-12-10
PR: Fix bad package dependency in docs/environment.yml

darkvertex:
Should fix the ReadTheDocs builds that have been broken for a while:
https://readthedocs.org/projects/qtawesome/builds/8159001/


ccordoba12:
Thanks!

Issue #107, opened 2019-01-18
Some FontAwesome icons do not match


jnsebgosselin:
Here is an minimal working example showing three problematic icons:

```python
import sys
import qtawesome as qta
from PyQt5.QtWidgets import (QApplication, QWidget, QGridLayout,
                             QToolButton, QLabel)

app = QApplication(sys.argv)

window = QWidget()
layout = QGridLayout(window)
names = ['fa5s.grip-horizontal', 'fa5s.file-export', 'fa5s.file-import']
for i, name in enumerate(names):
    qicon = qta.icon(name, scale_factor=1.3)
    btn = QToolButton()
    btn.setIcon(qicon)
    btn.setIconSize(QSize(32, 32))
    layout.addWidget(btn, i, 0)
    layout.addWidget(QLabel(name), i, 1)
window.show()

sys.exit(app.exec_())
```

![image](https://user-images.githubusercontent.com/10170372/51407136-07347c00-1b29-11e9-9acb-020400c6f1da.png)

https://fontawesome.com/icons/grip-horizontal?style=solid
https://fontawesome.com/icons/file-export?style=solid
https://fontawesome.com/icons/file-import?style=solid

ccordoba12:
Thanks for reporting @jnsebgosselin. Quick question: does this happen if you downgrade to **0.5.3**?

jnsebgosselin:
This happens with version 0.5.3 and 0.5.5.

darkvertex:
Hello there! :wave: 

Upon closer inspection, it would appear it might be an issue specific to Windows. I checked your icon names that appeared wrongly for you:

* [grip-horizontal](https://github.com/spyder-ide/qtawesome/blob/d2cbc1dc0555f6575205f77d21acb8d0cb917fed/qtawesome/fonts/fontawesome5-solid-webfont-charmap.json#L387) is `f58d` and matches with https://fontawesome.com/icons/grip-horizontal?style=solid

* [file-export](https://github.com/spyder-ide/qtawesome/blob/d2cbc1dc0555f6575205f77d21acb8d0cb917fed/qtawesome/fonts/fontawesome5-solid-webfont-charmap.json#L307
) is `f56e` and matches with https://fontawesome.com/icons/file-export?style=solid

[...]

PR #110, opened 2019-01-24
PR: Update update_fa5 command to rename internal font names during update

darkvertex:
In an effort to prevent further mistakes, I took the core code from https://github.com/chrissimpkins/fontname.py and made it into a utility function that then renames "regular" and "solid" flavours of FA5 to prevent issues like #107 .

This means that the command technically introduces a requirement for the "fonttools" module. Since it is a "dev" command I did not include it in the setup.py's `install_requires` but I did a lot of reading and it seems very tricky to declare the dependency for the command only. There's an `extra_requires` thing but that only works for package "entry_points" (which we don't use.) If you know a better way, I'm all ears.

For now, it'll raise an exception if the import fails.

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -10,11 +10,8 @@ import zipfile
 
 try:
     from fontTools import ttLib
-    # from fontTools.misc.py23 import tounicode, unicode
 except ImportError:
-    sys.exit(
-        "This special command requires the module 'fonttools': "
-        "https://github.com/fonttools/fonttools/")
+    ttLib = None
 
 try:
     # Python 2
@@ -197,9 +194,16 @@ class UpdateFA5Command(distutils.cmd.Command):
 
             # Fix to prevent repeated font names:
             if style in ('regular', 'solid'):
-                newName = str("Font Awesome 5 Free %s") % style.title()
-                self.__print('Renaming font to "%s" in: %s' % (newName, font_path))
-                rename_font(font_path, newName)
+                new_name = str("Font Awesome 5 Free %s") % style.title()
+                self.__print('Renaming font to "%s" in: %s' % (new_name, font_path))
+                if ttlib is not None:
+                    rename_font(font_path, new_name)
+                else:
+                    sys.exit(
+                        "This special command requires the module 'fonttools': "
+                        "https://github.com/fonttools/fonttools/")
+
+                # Reread the data since we just edited the font file:
                 with open(font_path, 'rb') as f:
                     data = f.read()
                     files[style] = data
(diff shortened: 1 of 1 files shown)


setupbase.py
```diff
@@ -129,13 +189,24 @@ def run(self):
...
-            with open(fontPath, 'w+') as f:
+            self.__print('Dumping updated "%s" font: %s' % (style, font_path))
+            with open(font_path, 'w+') as f:
                 f.write(data)
 
+            # Fix to prevent repeated font names:
+            if style in ('regular', 'solid'):
+                newName = str("Font Awesome 5 Free %s") % style.title()
+                self.__print('Renaming font to "%s" in: %s' % (newName, font_path))
+                rename_font(font_path, newName)
```
ccordoba12: Please change this line to

```python
if ttlib is not None:
    rename_font(font_path, new_name)
else:
    sys.exit(
        "This special command requires the module 'fonttools': "
        "https://github.com/fonttools/fonttools/")
```

setupbase.py
```diff
@@ -129,13 +189,24 @@ def run(self):
...
-            hashes[style] = hashlib.md5(data).hexdigest()
-            self.__print('Dumping updated "%s" font: %s' % (style, fontPath))
-            with open(fontPath, 'w+') as f:
+            self.__print('Dumping updated "%s" font: %s' % (style, font_path))
+            with open(font_path, 'w+') as f:
                 f.write(data)
 
+            # Fix to prevent repeated font names:
+            if style in ('regular', 'solid'):
+                newName = str("Font Awesome 5 Free %s") % style.title()
```
ccordoba12: Please change `newName` here to `new_name`.

darkvertex: Woops! Where I work our style guide is to use camelCase. I know this project uses snake_case and I made a mental effort to make it so, but I guess habit got the best of me! hehe. I'll edit it.

setupbase.py
```diff
@@ -1,10 +1,21 @@
...
 import hashlib
 import zipfile
+
+try:
+    from fontTools import ttLib
+    # from fontTools.misc.py23 import tounicode, unicode
+except ImportError:
+    sys.exit(
+        "This special command requires the module 'fonttools': "
+        "https://github.com/fonttools/fonttools/")
```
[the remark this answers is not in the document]
darkvertex: Ah! True. Thanks for the tip.

darkvertex:
@ccordoba12 I addressed your feedback. Tests are clearing now. Thanks for the review.

setupbase.py
```diff
@@ -1,10 +1,18 @@
 # -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
ccordoba12: Why is this line necessary?

darkvertex: I'm not 100% sure to be honest. I had kept it when copying code from: https://github.com/chrissimpkins/fontname.py/blob/master/fontname.py#L20
I did very minimal editing to make it usable as a helper function, but the bulk of the logic is as-is.

I just tried running without it and it seems fine, so I guess it's not really needed.

PR #114, opened 2019-03-06
Provide CharIconEngine.isNull() returning False

### LINKED ISSUE #113: Corrupted icons with PyQt 5.12
Hey, I had some issues with a corrupted windows profile, and after restoring it for some reason I can't get QtAwesome to show icons anymore. I did do a pip uninstall and install of qtawesome again. Normal QIcon are fine. Not sure on the technical side of this or how it could at all be connected, but I can't figure it out for the life of me. 

image

Icons are being added normally as 
```py
import qtawesome as qta
qta.icon("fa5s.mountain") 
```

ddemidov:
The default implementation of `QIconEngine.isNull()` returns True, which means other methods are never even attempted in PyQT 5.12.

Fixes #113

### DIFF
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -162,6 +162,9 @@ class CharIconEngine(QIconEngine):
         self.paint(QPainter(pm), QRect(QPoint(0, 0), size), mode, state)
         return pm
 
+    def isNull(self):
+        return False
+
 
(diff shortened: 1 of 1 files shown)


ddemidov:
This does not fix the issue. Tested on the wrong machine (with pyqt 5.11). Sorry about the noise.

PR #115, opened 2019-03-07
PR: Keep icons alive in cache

### LINKED ISSUE #113: Corrupted icons with PyQt 5.12

ddemidov:
Fixes #113

Looks like icons are stored by reference in qt 5.12, so we need to keep
them alive somewhere.

Not sure though if the icon cache should reside in the library, or in user code.
The advantage of this PR is that it works without any changes to the user code.

### DIFF
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -185,6 +185,7 @@ class IconicFont(QObject):
         self.painters = {}
         self.fontname = {}
         self.charmap = {}
+        self.icon_cache = {}
         for fargs in args:
             self.load_font(*fargs)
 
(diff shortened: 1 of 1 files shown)


stonebig:
I confirm the patch works ! Thank ou so much for the inquiry ### @ddemidov !

before (Spyder 2 theme)
image

now (Spyder 3 theme)
image

ccordoba12:
Thanks a lot @ddemidov for your help and @stonebig for the confirmation!

These are really great news indeed! I'll release **0.5.7** right away with this fix.

PR #117, opened 2019-05-28
PR: Add rotated, hflip, and vflip options

### LINKED ISSUE #112: Rotate icons by static angle i.e. 90, 180 etc
Is there a way I can statically rotate icons 90 or 180 degrees? The in html version for the material design font states that CSS helper classes can be used to rotate an icon.

wa-dev:
I attempted to address #112. The added behavior mimic the helper CSS classes detailed on the material design icons usage page (https://cdn.materialdesignicons.com/3.0.39/)

Fixes #112.

### DIFF
--- a/example.py
+++ b/example.py
@@ -26,6 +26,17 @@ class AwesomeExample(QtWidgets.QDialog):
         mdi_icon = qta.icon('mdi.access-point-network')
         mdi_button = QtWidgets.QPushButton(mdi_icon, 'Material Icons!')
 
+        # rotated
+        rot_icon = qta.icon('mdi.access-point-network', rotated=45)
+        rot_button = QtWidgets.QPushButton(rot_icon, 'Rotated Icons!')
+
+        # hflip
+        hflip_icon = qta.icon('mdi.account-alert', hflip=True)
+        hflip_button = QtWidgets.QPushButton(hflip_icon, 'Horizontally Flipped Icons!')
+
+        # vflip
+        vflip_icon = qta.icon('mdi.account-alert', vflip=True)
+        vflip_button = QtWidgets.QPushButton(vflip_icon, 'Vertically Flipped Icons!')
         # Styling
         styling_icon = qta.icon('fa5s.music',
                                 active='fa5s.balance-scale',
@@ -87,6 +98,7 @@ class AwesomeExample(QtWidgets.QDialog):
         # Layout
         vbox = QtWidgets.QVBoxLayout()
         widgets = [fa_button, elusive_button, mdi_button, music_button,
+                   rot_button, hflip_button, vflip_button,
                    toggle_button, stack_button, saveall_button, spin_button,
                    pulse_button, stack_spin_button, label]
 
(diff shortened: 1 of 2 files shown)


ccordoba12:
@wa-dev, thanks a lot for your contribution!

@dalthviz, please help me to review this one to see if it's working as expected. Simply run `example.py` and check @wa-dev's additions to it.

goanpeca:
What happens when someone does

```python
vflip_icon = qta.icon('mdi.account-alert', vflip=True, hflip=True, rotated=80)
```
?

What do you believe should happen @dalthviz?

ccordoba12:
@wa-dev, what do you think about @goanpeca's comment?

dalthviz:
@goanpeca as it is now you can only rotate and do one kind of flip (vertical or horizontal). The precedence of the actions is the following:

1. `rotated`
2. `hflip`
3. `vflip`

In terms of what is the expected behaviour, I think that one option (although I don't know how much complexity it would add) is to have a parameter to set the order in which the operations should be done, maybe a list of constants something like: 
```python
transform_order=[ROTATE, V_FLIP, H_FLIP]
```

goanpeca:
>transform_order=[ROTATE, V_FLIP, H_FLIP]

That seems like too much. I was expecting to either raise an error if more than two are provided?

Or if we are going to limit it to just one, then check that and raise the corresponding error.

For the way the examples were set up it seems like we should restrict to one of the three options only and raise an error otherwise @wa-dev

wa-dev:
Given
```python
vflip_icon = qta.icon('mdi.account-alert', vflip=True, hflip=True, rotated=80)
```
The combination of `vflip=True, hflip=True` is equivalent to a 180 degree rotation.
So  `vflip=True, hflip=True, rotated=80` should be equivalent to a 260 degree rotation.

I would think that `hflip` and `vflip` should be independent and should be applied first--then `rotated`.

goanpeca:
Hmm I prefer as separate attrobutes in that case (as you already have them)

ccordoba12:
@goanpeca, so you agree with merging @wa-dev's work as it is?

wa-dev:
As it is `hflip` and `vflip` are mutually exclusive options. Should that be the case?

goanpeca:
@wa-dev 

>As it is hflip and vflip are mutually exclusive options. Should that be the case?

I think they should not be mutually exclusive, I guess with what you explained we could just allow all three of them

@ccordoba12 

>Hmm I prefer as separate attrobutes in that case (as you already have them)

We should probably make hflip and vflip not exclusive

wa-dev:
What about the order.. `vflip`, `hflip`, and then `rotated`?

goanpeca:
Sure! That order seems reasonable

ccordoba12:
@wa-dev, sorry, I didn't notice that you updated this PR.

@goanpeca, do you think this is ready?

goanpeca:
Yes @ccordoba12  this is good to go!

PR #123, opened 2019-10-02
PR: Add an icon browser

JasonGilholme:
This adds a utility to browse all the available icons from the current QtAwesome installation.  You can also copy the name and python code to instantiate the selected icon.

After installing, you can run it from the command line with `qta_icon_browser`


goanpeca:
Thanks for this @JasonGilholme ! This is a great addition to QtAwesome.

I will get back to you later today with a review!

Cheers

goanpeca:
Some comments:

image

- Could we add the copy options on right click on top of each item?
- Could we make add some shortcuts for the same actions? Maybe `enter` for one and `shift+enter` for the other?
- Could we make the search case insensitive?

Some other nitpicks that could go for another PR or if you have some time?

We could add a listwidget to list the Icon Groups with checkboxes (to select/unselect some icons)
We could improve the search so that it is fuzzy. You could use something like this:
https://github.com/spyder-ide/spyder/blob/master/spyder/utils/stringmatching.py#L50

We could add a small test that at least runs the app and closes it after a while with `pytest-qt` I can give you some pointers if you need.

Thanks again!

goanpeca:
We also need to update the README with a screenshot and details on how to use it :-)

JasonGilholme:
> * Could we add the copy options on right click on top of each item?
> * Could we make add some shortcuts for the same actions? Maybe `enter` for one and `shift+enter` for the other?
> * Could we make the search case insensitive?
> * Could we sort the items? I see `fa` at the start and at the end

I've addressed the above comments.

> Some other nitpicks that could go for another PR or if you have some time?
> 
> * We could add a listwidget to list the Icon Groups with checkboxes (to select/unselect some icons)
> * We could improve the search so that it is fuzzy. You could use something like this:
>   https://github.com/spyder-ide/spyder/blob/master/spyder/utils/stringmatching.py#L50

I'm happy for that to go in another MR - here's some other ideas for a more advanced version of this:

* Preview pane on the right hand side that shows a larger version of the icon
* Additional settings to control values for the various kwargs to `.icon()` - color, color_active etc
* Control for the python namespace - support `import qtawesome as qta` for example.

> We could add a small test that at least runs the app and closes it after a while with `pytest-qt` I can give you some pointers if you need.

If you've got something handy that'd be good.

> Thanks again!

No problem - it's a useful utility that was missing from my workflow. I'll update the README.txt on the next push too.

JasonGilholme:
Updated the readme and added a screenshot.  Squashed all the commits too.

goanpeca:
Something like this... did not test it.  You will need to update CI services to install `pytest-qt`

```python

"""
Tests for QtAwesome Icon Browser.
"""

# Third party imports
from qtpy.QtCore import QTimer

# Local imports
from qtawesome import IconBrowser


@pytest.fixture
def browser(qtbot):
    browser = IconBrowser()
    qtbot.add_widget(browser)
    browser.show()
    return browser


def test_browser(qtbot, browser):
    def close():
        browser.close()

    timer = QTimer()
    timer.timeout.connect(close)
    timer.setSingleShot(2000)
    timer.start()


if __name__ == "__main__":
    pytest.main()
```

qtawesome/icon_browser.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        indexes = self._listView.selectedIndexes()
+        if not indexes:
+            return
+
+        iconName = indexes[0].data()
+        if code is True:
+            content = "qtawesome.icon('%s')" % iconName
+        else:
+            content = "'%s'" % iconName
```
darkvertex: It's weird to me that we copy the icon name with quotes in it.

JasonGilholme: Sure - I can update the second case.  The first case will still exhibit the problem that you describe however.

darkvertex: Could just solve both cases altogether by not supporting the copying as "code".

I think people are smart enough to read the README and figure out how to get an icon object. It's not that obscure.

JasonGilholme: I think you're right - I've removed the option.  You can now only copy the name without any quotes around it.

JasonGilholme:
Hey @goanpeca - I've addressed the outstanding comment and added some qt tests for the browser window which pas on CI.

There seems to be an issue specifically with the python3.5 tests though - any ideas?

goanpeca:
> I've addressed the outstanding comment and added some qt tests for the browser window which pass CI.

Awesome, thanks for working on this!

>There seems to be an issue specifically with the python3.5 tests though - any ideas?

So I would say we could bump all to test 3.6, 3.7 only and maybe 3.8. Also I would rather use now github actions for all OS.

Any thoughts here @ccordoba12 ?

JasonGilholme:
So as to not delay this PR longer further, should I update the existing CI configs to drop python-3.5 and add python-3.7/8 . ??  Similar to what's going on in this PR:  https://github.com/spyder-ide/qtawesome/pull/132

goanpeca:
@JasonGilholme I created this PR https://github.com/spyder-ide/qtawesome/pull/135 which moves everything to use github actions, which makes things much easier on our side. 

It should be ready to merge soon. As soon as it is merged, please merge with master so we can merge this PR.

Thanks for the patience with this :-)

JasonGilholme:
Awesome.  Yeah, no problem.

JasonGilholme:
Updated to provide some simple font collection filtering.

goanpeca:
Cool, please merge with master so new CI can run. Thanks!

JasonGilholme:
Updated and passed tests.

goanpeca:
Thanks for this contribution! Merging

PR #160, opened 2020-12-07
PR: Perform a font id validation

iLLiCiTiT:
## Issue
- singleton instance does not check if it's fonts are still available in `QFontDatabase`

## Changes
- `IconicFont` object also store loaded font ids
- added validation callback when `_instance()` is called to check if font ids are still in `QFontDatabase`


ccordoba12 (approved):
Thanks @iLLiCiTiT for your contribution! Looks good to me.

PR #146, opened 2020-05-27
PR: Timer should belong to parent widget

listyque:
If parent widget gets deleted timer automatically deleted to, instead timer tries to update already deleted widget.

PR #164, opened 2021-06-07
Kill QTimer when spinner parent is destroyed

santerioksanen:
Hi,

First of all thanks for an excellent library!
I run into issues with the spinner: I deleted the parent element, but the QTimer seemed to be left calling Spin.update at regular intervals. This caused runtime errors (RuntimeError: Internal C++ object (IconWidget) already deleted.).

Fixed by listening to parents deleted signal, and stopping the timer upon this.

Please let me know if any more tests are required.

### DIFF
--- a/qtawesome/animation.py
+++ b/qtawesome/animation.py
@@ -7,6 +7,7 @@ class Spin:
         self.parent_widget = parent_widget
         self.interval, self.step = interval, step
         self.info = {}
+        self.timer = None
 
     def _update(self):
         if self.parent_widget in self.info:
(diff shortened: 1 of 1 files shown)


dalthviz:
Hi @santerioksanen thanks for your contribution! I'm not totally sure, but maybe this fix could be covered with the change proposed at https://github.com/spyder-ide/qtawesome/pull/146 ? Could you confirm us that please? Thanks!

Pinging @ccordoba12 just in case

santerioksanen:
@dalthviz, I can confirm #146 also fixes the same issue in a cleaner way. I'll close this PR.

dalthviz:
Thank you for the confirmation @santerioksanen ! We will then merge the other one and you will have the fix in an incoming release :+1:

PR #166, opened 2021-07-26
PR: Support alpha in colors

ccordoba12:
- This is required for https://github.com/spyder-ide/spyder/pull/15964.
- Docs and gif were updated to mention this new functionality.

Example: This is a red icon with an alpha of 120.

imagen


dalthviz (approved):
@ccordoba12 LGTM :+1:

ccordoba12:
Great, thanks @dalthviz! At the end I realized this is not enough to solve the problem we have in spyder-ide/spyder#15964, but it's a nice addition to QtAwesome.

PR #174, opened 2021-10-19
PR: Update Material Design Icons to version 6.3.95

kumattau:
This PR is for updating Material Design Icons to version 6.3.95.

Compared to previous bundled version 5.4.55,

* License is changed from SIL Open Font License to Apache License Version 2.0.
* The number of icons are increased from 5555 to 6395, but breaking changes (some icons are renamed or removed) are introduced in version 6.1.95 (https://github.com/Templarian/MaterialDesign/issues/5409).

I think this PR can be merged If the breaking changes are no problem.


dalthviz:
Hi @kumattau thank you for the work here!

Regarding the breaking changes we will need to review them carefully since probably applications using `QtAwesome` could break as for example Spyder (which is something that [happened before](https://github.com/spyder-ide/spyder/issues/13875)). So first, we need to plan better the introduction of font breaking changes which maybe should be reflected in the version of `QtAwesome` (then this should be included for a `QtAwesome` `2.0.0` release) or think if `QtAwesome` should help keeping some level of compatibility when doing the font updates.

What do you think @ccordoba12 ?

kumattau:
Hi, @dalthviz

> I will create "PR: adding 6.3.95" and "PR: updating 5.3.55 to 5.9.55" except for updating UPDATE.md.

I finished the updates including UPDATE.md. I force pushed 3 commits to this PR.

PR #179, opened 2021-10-22
PR: Change access of constants to scoped enum style for PyQt6

kumattau:
This PR is for changing access of constants to scoped enum style for PyQt6.
(I used https://github.com/spyder-ide/qtpy/issues/233#issuecomment-820964838 's  tools to change)

As https://github.com/spyder-ide/qtpy/issues/233#issuecomment-946751764 says

> all bindings from 5.12+ (they may be supported in older versions, I just didn't check) supported the scoped enums

in the case of PyQt 5.9.2 provided by the latest conda,
the constants contained in qtawesome can be accessed via scoped enums.

I think this PR can be merged if support of qtawesome is enough with PyQt 5.9.2 or later.

I tested scoped enums using https://gist.github.com/kumattau/fa9e75b4cea6ec3722d9960f5a3996c9.

### DIFF
--- a/qtawesome/icon_browser.py
+++ b/qtawesome/icon_browser.py
@@ -45,17 +45,17 @@ class IconBrowser(QtWidgets.QMainWindow):
 
         self._proxyModel = QtCore.QSortFilterProxyModel()
         self._proxyModel.setSourceModel(model)
-        self._proxyModel.setFilterCaseSensitivity(QtCore.Qt.CaseInsensitive)
+        self._proxyModel.setFilterCaseSensitivity(QtCore.Qt.CaseSensitivity.CaseInsensitive)
 
         self._listView = IconListView(self)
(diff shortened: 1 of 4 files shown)


CAM-Gerlach:
NB, seems like we decided to have QtPy promote scoped to unscoped enums in PyQt6 for compatibility purposes, in spyder-ide/qtpy#271

kumattau:
Hi @CAM-Gerlach

Thank you for info.
I will close this PR after I check the original qtawesome works with qtpy, which supports PyQt6.

kumattau:
> I will close this PR after I check the original qtawesome works with qtpy, which supports PyQt6.

I checked that the original example.py and qtabroswer with the #183 fix work fine with qtpy, which supports PyQt6.
Note that the #183 fix is not directly related to unscoped enums.
So, I close this PR.

Issue #189, opened 2021-12-06
Incorrect FontAwesome5 Version in docs and update FontAwesome5 to the latest release (5.15.4)


kumattau:
README.md says "FA 5.9.0 features 1,534 free icons in different styles" but FA 5.14.0 seems to be bundled.

I checked the version as follows:

* FA 5.9.0: there are diffs in font and charmap.

```
python setup.py update_fa5 --fa-version 5.9.0
git diff
--- a/qtawesome/fonts/fontawesome5-brands-webfont-charmap.json
+++ b/qtawesome/fonts/fontawesome5-brands-webfont-charmap.json
@@ -48,7 +48,6 @@
     "btc": "f15a",
     "buffer": "f837",
     "buromobelexperte": "f37f",
-    "buy-n-large": "f8a6",
     "buysellads": "f20d",
     "canadian-maple-leaf": "f785",
     "cc-amazon-pay": "f42d",
@@ -73,7 +72,6 @@
     "confluence": "f78d",
     "connectdevelop": "f20e",
     "contao": "f26d",
...
```

* FA 5.14.0: there is diff in font only

```
python setup.py update_fa5 --fa-version 5.14.0
git diff
...
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -33,8 +33,8 @@ SYSTEM_FONTS = False
 # MD5 Hashes for font files bundled with qtawesome:
 MD5_HASHES = {
     'fontawesome4.7-webfont.ttf': 'b06871f281fee6b241d60582ae9369b9',
-    'fontawesome5-regular-webfont.ttf': '808833867034fb67a4a86dd2155e195d',
-    'fontawesome5-solid-webfont.ttf': '139654bb0acaba6b00ae30d5faf3d02f',
+    'fontawesome5-regular-webfont.ttf': '67e860127d1df8f1a3006c83f755fac2',
+    'fontawesome5-solid-webfont.ttf': '1ee101efc765a4068b360a3897f47cd4',
     'fontawesome5-brands-webfont.ttf': '085b1dd8427dbeff10bd55410915a3f6',
     'elusiveicons-webfont.ttf': '207966b04c032d5b873fd595a211582e',
     'materialdesignicons5-webfont.ttf': 'b7d40e7ef80c1d4af6d94902af66e524',
```

Since the timestamp when the family is changed is embedded, the font checksum changes every time the command is executed, which is to be expected.

dalthviz:
[...]

PR #190, opened 2021-12-08
PR: Preserves timestamp, bboxes, and table order for faithful to original and reproducible font updating

kumattau:
This PR resolves the following in https://github.com/spyder-ide/qtawesome/issues/189#issuecomment-988809501

>> Since the timestamp when the family is changed is embedded, the font checksum changes every time the command is executed, which is to be expected.
>
> * implement reproducing processing for changing family

By recalcTimestamp=False, "modified" timestamp field is preserved and same processing to change family returns same binary.
Also, recalcBBoxes=False and reorderTables=False are added to suppress processing that may change the table and content of font.

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -29,7 +29,7 @@ def rename_font(font_path, font_name):
     Font renaming code originally from:
     https://github.com/chrissimpkins/fontname.py/blob/master/fontname.py
     """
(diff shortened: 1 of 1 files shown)


dalthviz (approved):
Thanks @kumattau ! LGTM :+1:

PR #197, opened 2021-12-19
PR: Fix unexpected corruption check except for bundled fonts

### LINKED ISSUE #196: Unexpected corruption check when qta.load_font() with same ttf_filename as bundled
qta.load_font() checks corruption using ttf_filename as a key.
So, when users execute qta.load_font() with same ttf_filename as bundled, 
the ttf binary is checked unexpectedly and raise exception if the ttf binary is different from the bundled.
This makes the use of newer font than bundled by users difficult.

```python
# -*- coding: utf-8 -*-
import sys
from qtpy import QtWidgets, QtCore
import qtawesome as qta

class AwesomeExample(QtWidgets.QDialog):

    def __init__(self):
        super().__init__()
        # same ttf_filename as bundled
        qta.load_font("new_msc", "codicon.ttf", "codicon-charmap.json", directory="./fonts/")

        icon = qta.icon("new_msc.verified-filled")
        button = QtWidgets.QPushButton(icon, "New Codicon Icon")

        grid = QtWidgets.QGridLayout()
        grid.addWidget(button, 0, 0)
        self.setLayout(grid)
        self.setWindowTitle('Awesome')
        self.show()

def main():
    app = QtWidgets.QApplication(sys.argv)
    QtCore.QTimer.singleShot(10000, app.exit)
    _ = AwesomeExample()
    sys.exit(app.exec_())

if __name__ == '__main__':
    main()
```

```
$ python test.py
Traceback (most recent call last):
  File "test.py", line 29, in <module>
    main()
  File "test.py", line 25, in main
    _ = AwesomeExample()
  File "test.py", line 11, in __init__
    qta.load_font("new_msc", "codicon.ttf", "codicon-charmap.json", directory="./fonts/")
  File "/home/.local/lib/python3.8/site-packages/qtawesome/__init__.py", line 216, in load_font
    return _instance().load_font(prefix, ttf_filename, charmap_filename, directory)
  File "/home/.local/lib/python3.8/site-packages/qtawesome/iconic_font.py", line 317, in load_font
    raise FontError(u"Font is corrupt at: '{0}'".format(
qtawesome.iconic_font.FontError: Font is corrupt at: './fonts/codicon.ttf'
```

kumattau:
Fixes #196

```SYSTEM_FONTS``` has been removed in iconic_fonts.py because it is no longer needed.

I moved ```MD5_HASHES``` to \_\_init__.py and defined ```BUNDLED_FONTS``` in \_\_init__.py,
but I don't know "_" (underscore) should be prefixed in these variables  as internal use.

### DIFF
--- a/qtawesome/__init__.py
+++ b/qtawesome/__init__.py
@@ -16,17 +16,69 @@ Font-Awesome and other iconic fonts for PyQt / PySide applications.
    set_defaults
 """
 
+import hashlib
+import os
+
 # Third party imports
 from qtpy import QtCore, QtWidgets, QtGui
 
 # Local imports
 from ._version import __version__, version_info
 from .animation import Pulse, Spin
-from .iconic_font import IconicFont, set_global_defaults
+from .iconic_font import IconicFont, set_global_defaults, FontError
+from .iconic_font import SYSTEM_FONTS as _SYSTEM_FONTS
 from .styles import dark, light
 
 # Constants
 _resource = { 'iconic': None }
+
+_BUNDLED_FONTS = (
+    ('fa',
+        'fontawesome4.7-webfont.ttf',
+        'fontawesome4.7-webfont-charmap.json'),
+    ('fa5',
+        'fontawesome5-regular-webfont.ttf',
+        'fontawesome5-regular-webfont-charmap.json'),
+    ('fa5s',
+        'fontawesome5-solid-webfont.ttf',
+        'fontawesome5-solid-webfont-charmap.json'),
+    ('fa5b',
+        'fontawesome5-brands-webfont.ttf',
+        'fontawesome5-brands-webfont-charmap.json'),
+    ('ei',
+        'elusiveicons-webfont.ttf',
+        'elusiveicons-webfont-charmap.json'),
+    ('mdi',
+        'materialdesignicons5-webfont.ttf',
+        'materialdesignicons5-webfont-charmap.json'),
+    ('mdi6',
+        'materialdesignicons6-webfont.ttf',
+        'materialdesignicons6-webfont-charmap.json'),
+    ('ph',
+        'phosphor.ttf',
+        'phosphor-charmap.json'),
+    ('ri',
+        'remixicon.ttf',
+        'remixicon-charmap.json'),
+    ('msc',
+        'codicon.ttf',
+        'codicon-charmap.json'),
+)
+
+
+# MD5 Hashes for font files bundled with qtawesome:
+_MD5_HASHES = {
+    'fontawesome4.7-webfont.ttf': 'b06871f281fee6b241d60582ae9369b9',
+    'fontawesome5-regular-webfont.ttf': '808833867034fb67a4a86dd2155e195d',
+    'fontawesome5-solid-webfont.ttf': '139654bb0acaba6b00ae30d5faf3d02f',
+    'fontawesome5-brands-webfont.ttf': '085b1dd8427dbeff10bd55410915a3f6',
+    'elusiveicons-webfont.ttf': '207966b04c032d5b873fd595a211582e',
+    'materialdesignicons5-webfont.ttf': 'b7d40e7ef80c1d4af6d94902af66e524',
+    'materialdesignicons6-webfont.ttf': '9a2f455e7cbce011368aee95d292613b',
+    'phosphor.ttf': '5b8dc57388b2d86243566b996cc3a789',
+    'remixicon.ttf': '888e61f04316f10bddfff7bee10c6dd0',
+    'codicon.ttf': 'ca2f9e22cee3a59156b3eded74d87784',
+}
 
 
 def has_valid_font_ids(inst):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: qtawesome/__init__.py [hunk 1]
```diff
@@ -60,28 +111,22 @@ def _instance():
...
+            if ttf_hash is not None:
+                hasher = hashlib.md5()
+                with open(os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                       "fonts", ttf_filename), 'rb') as f:
+                    content = f.read()
+                    hasher.update(content)
+                ttf_calculated_hash_code = hasher.hexdigest()
+                if ttf_calculated_hash_code != ttf_hash:
+                    raise FontError(u"Font is corrupt at: '{0}'".format(
+                                    os.path.join(directory, ttf_filename)))
```
dalthviz, line 127: I think `os.path.join(directory, ttf_filename)` needs to be replaced with `os.path.join(os.path.dirname(os.path.realpath(__file__)), "fonts", ttf_filename)`

setupbase.py
```diff
@@ -296,14 +296,14 @@ def run(self):
             self.__print('New hash is: %s' % md5)
 
         # Now it's time to patch "iconic_font.py":
```
dalthviz, line 298: Should be `__init__.py`

setupbase.py
```diff
@@ -210,7 +210,7 @@ def run(self):
             hashes[style] = hashlib.md5(data).hexdigest()
 
         # Now it's time to patch "iconic_font.py":
```
dalthviz, line 212: Should be `__init__.py`

dalthviz (commented):
Thanks @kumattau for spotting this and for your help implementing a fix! This is looking quite good :+1: I left some comments regarding missing places still referencing `iconic_font.py` (when patching the MD5 hashes in `setupbase.py`) and a missing change for the FontError message.

I think we should also update the `UPDATE.md` instructions (I think there are some references to `iconic_font.py` that should point to `__init__.py`).

setupbase.py
kumattau: Fixed at 0ae5d6bcf298f30747ad6086f19f25dfa08161c9

setupbase.py
kumattau: Fixed at 0ae5d6bcf298f30747ad6086f19f25dfa08161c9

qtawesome/__init__.py
kumattau: Fixed at 3faefb25886b536adcbaf832804f38bab1a9264a

kumattau:
@dalthviz thank you for review.
I fixed your suggestions and more.
Additionally, SYSTEM_FONTS was restored because debian packages are still required.
https://salsa.debian.org/python-team/packages/python-qtawesome/-/blob/master/debian/patches/0002-Use-system-fonts.patch
In my opinion, SYSTEM_FONTS is very ad hoc and should be removed, but the PR focus to fix the #196.

If okey, I will squash to single commit and force push it.

dalthviz:
Thanks for the update @kumattau ! This is looking good. If you want to do a squash before merging this go ahead :+1:

kumattau:
I forgot to check the following.

> I don't know "_" (underscore) should be prefixed in these variables as internal use.

Should I prefix "_" (underscore) in ```SYSTEM_FONTS```, ```BUNDLED_FONTS```

dalthviz:
>Should I prefix "_" (underscore) in SYSTEM_FONTS, BUNDLED_FONTS and MD5_HASHES in __init__.py as internal use ?

I think it makes sense to add the prefix `_` to `BUNDLED_FONTS` and `MD5_HASHES` @kumattau :+1: . In the case of `SYSTEM_FONT` since is something currently in use to patch QtAwesome in Debian I think it should stay as it is (although probably we need to check for a better approach for that, maybe in another issue).

> Maybe, I think FontError should not be prefixed same as IconicFont, set_global_defaults.

Not totally sure to understand what do you mean here, sorry 😅 (maybe something like not adding a `_` prefix to those classes/function and let their names as they are?)

kumattau:
> In the case of `SYSTEM_FONT` since is something currently in use to patch QtAwesome in Debian I think it should stay as it is (although probably we need to check for a better approach for that, maybe in another issue).

debian patches SYSTEM_FONT in ```iconic_font.py```, but not ```__init__.py```.
So, by the following code, "_" can be prefixed to SYSTEM_FONT of ```__init__.py```.

```python
from .iconic_font import SYSTEM_FONTS as _SYSTEM_FONTS
```

kumattau:
> Not totally sure to understand what do you mean here, sorry sweat_smile (maybe something like not adding a _ prefix to those classes/function and let their names as they are?)

I'm sorry for my poor explanation. I'll explain it again.

The FontError class in iconic_font.py, which has not been made public until now, will be exposed in the qtawesome namespace,
but I think this is no problem.

For example, classes and functions used by users, such as Spin in annimation.py, are exposed in the qtawesome namespace.

Similarly, FontError is a class used by users, so I think it is better to exposed in the qtawesome namespace.

By exposing, users can write the following code under qtawesome namespace.

````python
import qtawesome as qta

try:
    fa5_icon = qta.icon('fa5_flags')
except qta.FontError
    ...
```

dalthviz:
> debian patches SYSTEM_FONT in iconic_font.py, but not __init__.py.
> So, by the following code, "_" can be prefixed to SYSTEM_FONT of __init__.py.
> 
> from .iconic_font import SYSTEM_FONTS as _SYSTEM_FONTS

Sure we can do that alias :+1:

> I'm sorry for my poor explanation. I'll explain it again.
> 
> The FontError class in iconic_font.py, which has not been made public until now, will be exposed in the qtawesome namespace,
> but I think this is no problem.
> 
> For example, classes and functions used by users, such as Spin in annimation.py, are exposed in the qtawesome namespace.
> 
> Similarly, FontError is a class used by users, so I think it is better to exposed in the qtawesome namespace.
> 
> By exposing, users can write the following code under qtawesome namespace.
> 
> import qtawesome as qta
> 
> try:
>     fa5_icon = qta.icon('fa5_flags')
> except qta.FontError:
>     ...

Ah yes, no problem with that :+1:

kumattau:
@dalthviz,

My work has been finished.
I squashed to single commit and force pushed it.

dalthviz (approved):
Thanks @kumattau for the help with this! LGTM :+1:

PR #202, opened 2022-01-26
PR: Use QRawFont and draw PainterPath of Glyph for animation to prevent tremulous spinning icons

### LINKED ISSUE #39: Tremulous spinning icons
I noticed that all spinning icons do shake a little while rotating.
Is there a way to make it smooth as on the original web page?

kumattau:
Maybe #39 can be fixed using QRawFont and draw PainterPath of Glyph.

Compared to QFont, QRawFont is available without registering to font database,
So this approach has

Pros:
* no tremulous spinning
* no family name collision (font patch is not required)

Cons:
* cannot support color-font (PainterPath of Glyph does not have color)
* cannot support font() API (this requires QFont registered to font database)
* I tests it on ubuntu only. I cannot check this approach works on other platform or not

If qtawesome switches QFont to QRawFont, font patch in the update_font script ( #194 ) may not be needed.

How do you think, @dalthviz  ?

Note: current code is example, so it needs refactoring, optimization, etc.

Edit:

Fixes #39

### DIFF
--- a/example.py
+++ b/example.py
@@ -114,6 +114,14 @@ class AwesomeExample(QtWidgets.QDialog):
                                     {'scale_factor': 0.8}])
         saveall_button = QtWidgets.QPushButton(saveall, 'Stack, offset')
 
+        # Stack and offset icons, draw path
+        saveall_path = qta.icon('fa5.save', 'fa5.save',
+                           options=[{'scale_factor': 0.8,
+                                     'offset': (0.2, 0.2),
+                                     'color': 'gray', 'draw': 'path'},
+                                    {'scale_factor': 0.8, 'draw': 'path'}])
+        saveall_button_path = QtWidgets.QPushButton(saveall_path, 'Stack, offset, draw path')
+
         # Spin icons
         spin_button = QtWidgets.QPushButton(' Spinning icon')
         spin_icon = qta.icon('fa5s.spinner', color='red',
@@ -167,6 +175,7 @@ class AwesomeExample(QtWidgets.QDialog):
             pulse_button,
             stack_button,
             saveall_button,
+            saveall_button_path,
             stack_spin_button,
         ]
         other_widgets = [
--- a/example.py
+++ b/example.py
@@ -114,6 +114,14 @@ class AwesomeExample(QtWidgets.QDialog):
                                     {'scale_factor': 0.8}])
         saveall_button = QtWidgets.QPushButton(saveall, 'Stack, offset')
 
+        # Stack and offset icons, draw image
+        saveall_image = qta.icon('fa5.save', 'fa5.save',
+                           options=[{'scale_factor': 0.8,
+                                     'offset': (0.2, 0.2),
+                                     'color': 'gray', 'draw': 'image'},
+                                    {'scale_factor': 0.8, 'draw': 'image'}])
+        saveall_button_image = QtWidgets.QPushButton(saveall_image, 'Stack, offset, draw image')
+
         # Stack and offset icons, draw path
         saveall_path = qta.icon('fa5.save', 'fa5.save',
                            options=[{'scale_factor': 0.8,
@@ -175,6 +183,7 @@ class AwesomeExample(QtWidgets.QDialog):
             pulse_button,
             stack_button,
             saveall_button,
+            saveall_button_image,
             saveall_button_path,
             stack_spin_button,
         ]
--- a/example.py
+++ b/example.py
@@ -114,22 +114,6 @@ class AwesomeExample(QtWidgets.QDialog):
                                     {'scale_factor': 0.8}])
         saveall_button = QtWidgets.QPushButton(saveall, 'Stack, offset')
 
-        # Stack and offset icons, draw image
-        saveall_image = qta.icon('fa5.save', 'fa5.save',
-                           options=[{'scale_factor': 0.8,
-                                     'offset': (0.2, 0.2),
-                                     'color': 'gray', 'draw': 'image'},
-                                    {'scale_factor': 0.8, 'draw': 'image'}])
-        saveall_button_image = QtWidgets.QPushButton(saveall_image, 'Stack, offset, draw image')
-
-        # Stack and offset icons, draw path
-        saveall_path = qta.icon('fa5.save', 'fa5.save',
-                           options=[{'scale_factor': 0.8,
-                                     'offset': (0.2, 0.2),
-                                     'color': 'gray', 'draw': 'path'},
-                                    {'scale_factor': 0.8, 'draw': 'path'}])
-        saveall_button_path = QtWidgets.QPushButton(saveall_path, 'Stack, offset, draw path')
-
         # Spin icons
         spin_button = QtWidgets.QPushButton(' Spinning icon')
         spin_icon = qta.icon('fa5s.spinner', color='red',
@@ -183,8 +167,6 @@ class AwesomeExample(QtWidgets.QDialog):
             pulse_button,
             stack_button,
             saveall_button,
-            saveall_button_image,
-            saveall_button_path,
             stack_spin_button,
         ]
         other_widgets = [
@@ -198,20 +180,36 @@ class AwesomeExample(QtWidgets.QDialog):
 
         for idx, w in enumerate(styled_widgets):
             grid.addWidget(w, idx, 1)
-        
+
         for idx, w in enumerate(animated_widgets):
             grid.addWidget(w, idx + len(styled_widgets), 1)
-        
+
         for idx, w in enumerate(other_widgets):
             grid.addWidget(w, idx + len(styled_widgets) + len(animated_widgets), 1)
 
+        title = 'Awesome'
+        args = ' '.join(sys.argv[1:]).strip()
+        if args:
+            title += ' (' + args + ')'
+
         self.setLayout(grid)
-        self.setWindowTitle('Awesome')
+        self.setWindowTitle(title)
         self.setMinimumWidth(520)
         self.show()
 
 
 def main():
+
+    global_defaults = {}
+    for arg in sys.argv[1:]:
+        try:
+            key, val = arg.split('=', maxsplit=1)
+            global_defaults[key] = val
+        except:
+            pass
+    if global_defaults:
+        qta.set_global_defaults(**global_defaults)
+
     app = QtWidgets.QApplication(sys.argv)
 
     # Enable High DPI display with PyQt5
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -22,9 +22,10 @@ import shutil
 import warnings
 
 # Third party imports
-from qtpy.QtCore import QByteArray, QObject, QPoint, QRect, Qt
+from qtpy.QtCore import (QByteArray, QObject, QPoint, QRect, Qt,
+                         QSizeF, QMarginsF, QRectF)
 from qtpy.QtGui import (QColor, QFont, QFontDatabase, QIcon, QIconEngine,
-                        QPainter, QPixmap, QTransform, QPalette)
+                        QPainter, QPixmap, QTransform, QPalette, QRawFont)
 from qtpy.QtWidgets import QApplication
 
 # Linux packagers, please set this to True if you want to make qtawesome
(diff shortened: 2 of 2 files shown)


dalthviz:
Hi @kumattau thanks for checking into this! Testing this locally seems like this also works for Windows :+1:

example

Regarding the pros and cons maybe we could use `QRawFont` only when needed as you are doing here but still registering the font to still have access to the font API if needed?

Just in case what do you think @ccordoba12 ?

kumattau:
Thank you for testing @dalthviz

On windows, "Stack, offset" bled quite a bit. this is because QPainterPath does not have hinting.
So, I changed it to use QPainterPath only for animations to keep rendering quality.

On my ubuntu, example.py is here.
Could you compare "Stack, offset" and "Stack, offset, draw path" on Windows ?

Screenshot from 2022-01-29 00-45-56

The drawing method can be selected by "draw" option.

* path: do painter.drawPath()
* text: do painter.drawText()
* auto: do painter.drawPath when animation, otherwise do painter.drawText() (default)

kumattau:
@dalthviz, I finished the update. This PR is ready for review.

dalthviz:
Checking the `example.py` on Windows this is what I'm getting:

imagen

example.py
```diff
@@ -114,6 +114,14 @@ def __init__(self):
...
         saveall_button = QtWidgets.QPushButton(saveall, 'Stack, offset')
 
+        # Stack and offset icons, draw path
+        saveall_path = qta.icon('fa5.save', 'fa5.save',
+                           options=[{'scale_factor': 0.8,
+                                     'offset': (0.2, 0.2),
+                                     'color': 'gray', 'draw': 'path'},
+                                    {'scale_factor': 0.8, 'draw': 'path'}])
+        saveall_button_path = QtWidgets.QPushButton(saveall_path, 'Stack, offset, draw path')
+
```
dalthviz, line 124: I think we should add this to the docs and probably update the GIF for `example.py`. Maybe adding to the docs some info regarding the new `draw` option

qtawesome/iconic_font.py
```diff
@@ -173,7 +173,47 @@ def _paint_icon(self, iconic, painter, rect, mode, state, options):
...
 
-        painter.drawText(rect, int(Qt.AlignCenter | Qt.AlignVCenter), char)
+        draw = options.get('draw', 'auto')
+        if draw == 'text':
+            draw_path = False
+        elif draw == 'path':
+            draw_path = True
+        else:
+            # Use QPainterPath when animation
+            # to fix tremulous spinning icons #39
```
dalthviz, line 183: Improve a little bit the comment here
```suggestion
            # Use QPainterPath when setting an animation
            # to fix tremulous spinning icons.
            # See #39
```

qtawesome/iconic_font.py
```diff
@@ -270,14 +312,23 @@ def hook(obj):
         # Load font
         if QApplication.instance() is not None:
             with open(os.path.join(directory, ttf_filename), 'rb') as font_data:
-                id_ = QFontDatabase.addApplicationFontFromData(QByteArray(font_data.read()))
+                data = font_data.read()
+                id_ = QFontDatabase.addApplicationFontFromData(data)
+                # QRawFont() requires pixelSize, but it is unknown at this time.
+                # So, first create QRowfont with 2048, which is the size of
+                # almost TrueType fonts and reset actual size after that.
+                # Note: When TrueType fonts, pixelSize can be specified up to 16384.
```
dalthviz, line 320: A little improvement to the comment here:
```suggestion
                # `QRawFont()` requires a `pixelSize` value, but it is unknown at this point.
                # To workaround that, first we create a `QRawFont` with 2048 as `pixelSize` 
                # (usual size for TrueType fonts) and reset to the actual size after that.
                # Note: For TrueType fonts, `pixelSize` can be specified up to 16384.
```

dalthviz (commented):
Thanks @kumattau for working on this! I left some suggestions regarding the comments in the code and also maybe we need to also update the docs (at least the example code in this PR) but otherwise this LGTM :+1:

kumattau:
Thank you for review @dalthviz, I updated comment.
I've come up with a better way to do it and will experiment with it. Please wait a little longer.

kumattau:
Hi, @dalthviz 

> I've come up with a better way to do it and will experiment with it.

I try to drawImage a glyph of QRawfont.
On my ubuntu, the drawing quality is same as original drawText with QFont.
Could you try example.py on Windows and feedback me ?

If the drawImage is okey, qtawesome does not require for QFont except for font() API.
I think I want not to patch fonts if possible, so, it is the first step of the use of patch-free font.

dalthviz:
Hi again @kumattau, this is what I'm seeing on Windows with the latest changes:

imagen

So, indeed the icon drawing locks cleaner :+1: 

However seems like the [new code is not compatible with Qt <5.15](https://forum.qt.io/topic/121460/qimage-setalphachannel-obsolete-or-not/3). The tests are failling here since it our CI uses Qt 5.9.7 and locally I tried first testing with an env that had Qt 5.12.2 and the same error appeared. Could be possible to have a fallback in case `QImage.setAlphaChannel` doesn't exists in the Qt/PyQt/PySide version installed?

kumattau:
@dalthviz 

> Could be possible to have a fallback in case QImage.setAlphaChannel doesn't exists in the Qt/PyQt/PySide version installed?

Yes. I added fallback code and pushed it.
I will try to refactor the code and add three draw options (path, image, text).

kumattau:
Hi, @dalthviz 

> cannot support color-font (PainterPath of Glyph does not have color)

I found another issue.
QRawFont requires that the font has glyph.
So, for example, NotoColorEmoji.ttf, which does not have glyphs, cannot be rendered using QRawFont.

https://github.com/fonttools/fonttools/issues/949#issuecomment-299148957

qtawesome cannot switch the drawing method from `drawPath()` to the original `drawText()` if the font does not have glyphs.
Because QRawFont have `bool supportsCharacter(char)` method to check if the font has a glyph that corresponds to the given character, but this always returns True on ubuntu.

And the following error message output to the console if a glyph is missing, but this message cannot stopped.

> load glyph failed err=24

Should qtawesome support a font which does not have glyphs ?

dalthviz:
Hi again @kumattau !

> Should qtawesome support a font which does not have glyphs ?

I would say yes. Probably we need to find a reliable way to detect if the font does not have the needed glyphs, show some sort of warning letting users know that we were unable to use the specified config to draw the icon and then fallback to use the `drawText`/``text` option.

Maybe there are other ways to check for the character support more reliable than `QRawFont.supportsCharacter` ?

kumattau:
Hi @dalthviz 

> I would say yes.

I got it.

> I would say yes. Probably we need to find a reliable way to detect if the font does not have the needed glyphs, show some sort of warning letting users know that we were unable to use the specified config to draw the icon and then fallback to use the drawText/``text` option.

I think fallback is possible.
If glyph is missing, the PainterPath is empty, so I can implement the fallback to drawText when PainterPath is emtpy.
The biggest issue at the moment is that I can't stop the following error messages.

> load glyph failed err=24

It is quite annoying because it is always output when glyph is missing.

I will do some more research on this matter.

dalthviz:
Hi again @kumattau , hope everything is going well! Since we will try to release v1.2.0 soon, just to know if we can make this part of the release a couple of questions: Did you find any other info that could help with the missing elements here? and also, would you like to finish this or is okay for me or someone else to try finish it? Let us know!

kumattau:
@dalthviz Thank you for follow-up and I'm very sorry not to response.

> I will do some more research on this matter.

The investigation has not progressed but I will reboot this PR soon (in this week or next week).
If my schedule is not suite the release, I'm no problem for you to try finish it at all.

dalthviz:
Thanks for the prompt response @kumattau ! If you still want to finish this at some point then no problem, happy to hear that :)

It would be nice to merge this for the next release v1.2.0 (which will be done probably mid-next week) but do not worry about the time, just knowing that this could be finished in the future and you are willing to keep checking on it are really good news :+1:

kumattau:
@dalthviz 

I rebased master and fixed "load glyph failed err=24" problem by fall backing draw=text
The draw option accepts text/path/glyphrun/image. If not specified, draw=path is used for non-animation and draw=text is used for animation.
example.py accepts global_default options (key=val format) for test.

In my linux, it is better that draw=image is used for non-animation and draw=text is used for animation.
Could you please try it in windows ?

kumattau:
Looking to spin, I still think draw=path would be better when animation.

dalthviz:
Thanks for the quick update @kumattau ! Doing some testing on Windows:

### Default options (no args) - `python example.py`

default

### `draw=image` option - `python example.py draw=image`

draw_image

### `draw=path` option - `python example.py draw=path`

draw_path

### `draw=text` option - `python example.py draw=image`

draw_text

Checking then seems like the `draw=path` should be the default when animation is requested and `draw=image` when not

qtawesome/iconic_font.py
```diff
@@ -178,45 +179,94 @@ def _paint_icon(self, iconic, painter, rect, mode, state, options):
...
 
         painter.setOpacity(options.get('opacity', 1.0))
 
-        painter.drawText(rect, int(Qt.AlignCenter | Qt.AlignVCenter), char)
+        draw = options.get('draw')
+        if draw not in ('text', 'path', 'glyphrun', 'image'):
+            # Use QPainterPath when setting an animation
+            # to fix tremulous spinning icons.
+            # See #39
+            draw = 'path' if animation is not None else 'text'
```
dalthviz, line 207: Following our findings I think here we should change `text` for `image`:

```suggestion
            draw = 'path' if animation is not None else 'image'
```

dalthviz:
Besides the suggestion above I think the only missing element is updating the `example.py` gif and documentation (adding the new draw option and also the possiblity to pass options to the `example.py` script) but I can take care of that in a follow-up PR before doing the release :+1:

qtawesome/iconic_font.py
```diff
@@ -178,45 +179,94 @@ def _paint_icon(self, iconic, painter, rect, mode, state, options):
...
+        painter.setTransform(transform, True)
 
         painter.setOpacity(options.get('opacity', 1.0))
 
-        painter.drawText(rect, int(Qt.AlignCenter | Qt.AlignVCenter), char)
+        draw = options.get('draw')
+        if draw not in ('text', 'path', 'glyphrun', 'image'):
+            # Use QPainterPath when setting an animation
+            # to fix tremulous spinning icons.
+            # See #39
```
dalthviz, line 206: Let change the comment to follow other comments pointing to issues:
```suggestion
            # See spyder-ide/qtawesome#39
```

qtawesome/iconic_font.py
kumattau: Looking at Material Design and Rotate, image seems not to be good.

kumattau:
@dalthviz 

How about draw=glyphrun on windows ?

On linux, The quality of draw=glyphrun seems to be same as draw=text.

`glyphrun` can use rawfont (in other words, does not requires the installation of font) but cannot render emoji (without glyph).

qtawesome/iconic_font.py
kumattau: As https://github.com/spyder-ide/qtawesome/pull/202#issuecomment-1279741225, I think `text` or `glyphrun` may be  better for non-animation on linux and windows.

dalthviz: You are right! I missed checking that and just focused on the remix icon that looked sharper with `draw=image` 😅

dalthviz:
Checking `draw=glyphrun` on Windows I got this:

image

Looking more in detail the different draw options I would say that `glyphrun` looks quite similar to `text` and also as you said for some fonts using `image` doesn't look quite good, so lets leave the default draw option to be `text` for non-animated icons :+1:

Thanks again @kumattau for the work here and checking all the draw options! I think this is ready to be merged then, right? Or maybe do you want to check something else on Windows? If is ready to be merged just mark the PR as `Ready for review` and I will merge it :+1:

kumattau:
@dalthviz 

My work has been finished.
The commits keep not to be squashed.
If you want to simplify the PR, please squash and merge it.

dalthviz (approved):
Thanks for all the work done here @kumattau !

PR #203, opened 2022-02-28
PR: Disable font hinting to mitigate tremulous spinning to some extent

kumattau:
Hi @dalthviz.

This PR is to mitigate #39 to some extent.

As https://github.com/spyder-ide/qtawesome/pull/202#issuecomment-1031631561, the method using QRawFont and QPainterPath does not support all fonts supported by QFont.
So, I think the method is optional.

Even if QFont is used, by disabling font hinting, the tremulous spinning seems to be mitigated to some extent.
So, I think it is better that the font hinting is disabled as default when animation.

left: default (enable) hinting, right: disable hinting
Animation

### DIFF
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -140,7 +140,12 @@ class CharIconPainter:
         if animation is not None:
             animation.setup(self, painter, rect)
 
-        painter.setFont(iconic.font(prefix, draw_size))
+        font = iconic.font(prefix, draw_size)
+        # Disable font hinting to mitigate tremulous spinning to some extent
+        # See spyder-ide/qtawesome#39
+        if animation is not None:
+            font.setHintingPreference(QFont.PreferNoHinting)
+        painter.setFont(font)
         if 'offset' in options:
             rect = QRect(rect)
             rect.translate(round(options['offset'][0] * rect.width()),


qtawesome/iconic_font.py
```diff
@@ -140,7 +140,10 @@ def _paint_icon(self, iconic, painter, rect, mode, state, options):
         if animation is not None:
             animation.setup(self, painter, rect)
 
-        painter.setFont(iconic.font(prefix, draw_size))
+        font = iconic.font(prefix, draw_size)
+        if animation is not None:
```
dalthviz, line 144: Could you add a comment here referencing the issue?

dalthviz (commented):
Thanks @kumattau for keep checking into this! I left a comment to add a reference to the issue here. Otherwise this LGTM :+1:

qtawesome/iconic_font.py
kumattau: I added the comment "Disable font hinting to mitigate tremulous spinning to some extent".
Is it as expected?

kumattau: I added the comment "Disable font hinting to mitigate tremulous spinning to some extent".
Is it as expected?

dalthviz: That sounds good :+1: it's just missing the addition of something like `See spyder-ide/qtawesome#39` at the end

kumattau: I added more "See spyder-ide/qtawesome#39" at the end.

dalthviz (approved):
Thanks @kumattau !

PR #232, opened 2023-04-12
PR: Remove usage of deprecated codecov package and move to GitHub Action variant

dalthviz:
Using the codecov package can causes CI to fail with `pip` since the `codecov` package was removed from PyPI. The error from CI is the following:

```
ERROR: Could not find a version that satisfies the requirement codecov (from versions: none)
ERROR: No matching distribution found for codecov
```

Example Spyder CI run where this happens: https://github.com/spyder-ide/spyder/actions/runs/4681680400/jobs/8294575413?pr=20421#step:10:420

Remove codecov from requirements and use Codecov GitHub Action instead to upload coverage reports


ccordoba12 (approved):
Thanks @dalthviz!

PR #233, opened 2023-05-04
PR: qta-browser update

pedromorgan:
Some `qta-browser` updates to make gui nicer. 

- A toolbar at top with actions
- make stuff persist with qsettings

**Hidden Motive**

- End goal is to "embed as widget in app" and let clients decide their own icons and colors ;-)

image


dalthviz:
Hi @pedromorgan thank you for your interest in improving the icon browser! I think the idea behind the changes makes sense but just in case, what do you think @ccordoba12 ?

Also, if this go forward I think we should update the screenshot for the icon browser and maybe expand a little bit the [README section about it](https://github.com/spyder-ide/qtawesome#icon-browser) (maybe including info about the settings that are persisted between launches?) and of course fix the failing tests :)

Thank you again for your interest in improving QtAwesome @pedromorgan !

pedromorgan:
hi @dalthviz, thanks for reply... and an updates patch

-  fixed the tests 
- removed settings, cos its not really necessary, as before

ccordoba12:
I looks good to me too, thanks @pedromorgan for your contribution!

pedromorgan:
Just updated screenshot 

Ready for merge.. TIA

pedromorgan:
@dalthviz @ccordoba12 How does this review work?

dalthviz:
Hi @pedromorgan sorry for the late response! Just approved the run of the CI to check that the tests are passing (and seems like everything is green now 🎉 ) and now we will give this a more throughful review :+1:

Just in case, after a quick check, looks like maybe the only couple of things I would say needs to be changed are:

* The icon of the window - Although pandas are nice, maybe using the `fa5s.icons` or `fa5b.font-awesome` icon could be better:

image

image

* Maybe the number of default columns (maybe to 6 or 8), since from the screenshot 10 seems like to much.

Anyhow, I would check this locally too as part of the review and maybe then the things above could not be so interesting to do or maybe some other suggestions could appear.

pedromorgan:
@dalthviz  fixed the icon...

Re columns, am using 30 across on a big screen.. the main reason I hacked in the first place..

dalthviz:
> Re columns, am using 30 across on a big screen.. the main reason I hacked in the first place..

Checked locally and actually I think the default columns number is okay :+1: However, the window minimum size is kind of big (at least for my screen) so probably we will need to reduce it a little bit. Locally I'm seeing:

 browser

I will leave some suggestions as a review to change the minimum size and some other little nitpicks (like the clear icon and maybe using tooltips instead of the addition of the `ToolBarGroup` class/instances as titles)

pedromorgan:
@dalthviz Done.. Fixed to minimum screen size ;-)))))))))))))

qtawesome/icon_browser.py
```diff
@@ -260,6 +311,31 @@ def data(self, index, role):
         return super().data(index, role)
 
 
+class ToolBarGroup(QtWidgets.QWidget):
```
dalthviz, line 314: Regarding this custom widget to add titles, what do you think if instead we just add tooltips to the relevant widgets so people can check/understand what they do?

So for example for the line edit that filters the icons:

filter

For the text for each element, it could be something like:

* Font combobox: `Select fonts prefix which icons will be included in the filtering`
* Filter line edit: `Filter icons by name`
* Selected icon line edit: `Full identifier of the currently selected icon`
* Copy button: `Copy selected icon full identifier to the clipboard`
* Theme combobox: `Select color palette for the icons and the icon browser`
* Columns combobox: `Select number of columns the icons list is showing`

qtawesome/tests/test_icon_browser.py
```diff
@@ -56,24 +56,34 @@ def test_copy(qtbot, browser):
...
 
-    assert "google" in clipboard.text()
+    assert "penguin" in clipboard.text()
 
-
-def test_filter(qtbot, browser):
+def test_filter_success(qtbot, browser):
     """
-    Ensure the filter UX works
+    What happens if a `penguin`
```
dalthviz, line 63: Lets rewrite this to:

```suggestion
    Ensure the filter UX works when searching for `penguin`
```

qtawesome/tests/test_icon_browser.py
```diff
@@ -56,24 +56,34 @@ def test_copy(qtbot, browser):
...
+    # Enter a search term and click
+    qtbot.keyClicks(browser._lineEditFilter, 'penguin')
+    qtbot.keyPress(browser._lineEditFilter, QtCore.Qt.Key_Enter)
+
+    filteredRowCount = browser._listView.model().rowCount()
+    assert filteredRowCount > 1
 
+def test_filter_fail(qtbot, browser):
+    """
+    What happens if a `not a penguin-penguin`
```
dalthviz, line 77: Lets rewrite this to be:

```suggestion
    Ensure the filter doesn't show results if the filter text doesn't mach any icon
```

dalthviz (commented):
Thank you again @pedromorgan for the work here! Left some initial comments regarding the widgets properties, move the list with possible number of columns to be a constant and also a way to simplify the code to clear the filter line edit.

Probably the biggest change that I propose in this review is to drop the usage of the custom class `ToolBarGroup` and just set tooltips for the different widgets, but let me know what do you think!

ccordoba12:
@pedromorgan, may I ask: why did you decide to close this pull request? We can help you finish it, if that's ok for you.

pedromorgan:
Because I got better things to do than fix some review on code style etc.  Also this is the first time in the discussion that a lot of the topics are brought up.

The time spent doing the review, could have been spent fixing the code itself.

So am not going to waste my time.. TA

pedromorgan:
I understand your role now.. A pointless job..

- So instead of actually doing things,
-  instead telling people to do things..
- instead of doing it themselves...

AFAIK..
- u should have merges it in your own branch
- clean it up to "YOUR" taste
- and then commit to main and job done..
- fingers crossed and job done

This explains a lot why foss project stall..
- and ultimately fork
- over a trivial issue

pedromorgan:
@dalthviz has spent more time revieing my code, that actually correct the errors within...

There is a weed in the garden
- I propose to remove it...
- No wait. .need to check first with code review

dalthviz:
I'm sorry @pedromorgan if the way I reviewed the PR frustated you. Just until yestarday I was able to make time to give a full review and it ended up having more changes/suggestions than I initially though. I still think that the work here is worthy to be finished and merged so if you are okay I can help finish the PR so it can be merged.

Again, truly sorry if the review process was not what you expected to be. By no means I wanted you to feel bad or frustated with this PR.

pedromorgan:
@derek-rein Why not just merge this, and go and fix it.. TIA.. ;-))))

- U spent more time on review, instead of taking the "kinda way i see it".. to fit your review ideals
- instead of merging it,, and then fixing it... even though my version is better than original > its now ever better..

Move faster please... and let ideas flow... Think more like a carpenter and wood rather than code..

PR #238, opened 2023-06-21
PR: Provide icon size when creating an `IconWidget`

StSav012:
I suggest adding a `size` keyword-only argument to the constructor of `IconWidget` to avoid extra `setIconSize` and `update` calls.

This way,
```python
icon_widget: qta.IconWidget = qta.IconWidget('mdi.web', color='blue')
icon_widget.setIconSize(QtCore.QSize(32, 32))
icon_widget.update()
```
turns into
```python
icon_widget: qta.IconWidget = qta.IconWidget('mdi.web', color='blue', size=QtCore.QSize(32, 32))
```

---

And I fixed a couple of typos in the docstring of the class.

### DIFF
--- a/example.py
+++ b/example.py
@@ -96,7 +96,7 @@ class AwesomeExample(QtWidgets.QDialog):
         lo.addWidget(iconwidget)
         lo.addWidget(QtWidgets.QLabel('IconWidget'))
         iconwidgetholder.setLayout(lo)
-        iconwidget2 = qta.IconWidget('mdi.web', color='blue')
+        iconwidget2 = qta.IconWidget('mdi.web', color='blue', size=QtCore.QSize(24, 24))
 
         # Icon drawn with the `image` option
         drawn_image_icon = qta.icon('ri.truck-fill',
--- a/example.py
+++ b/example.py
@@ -96,7 +96,7 @@ class AwesomeExample(QtWidgets.QDialog):
         lo.addWidget(iconwidget)
         lo.addWidget(QtWidgets.QLabel('IconWidget'))
         iconwidgetholder.setLayout(lo)
-        iconwidget2 = qta.IconWidget('mdi.web', color='blue', size=QtCore.QSize(24, 24))
+        iconwidget2 = qta.IconWidget('mdi.web', color='blue', size=QtCore.QSize(16, 16))
 
         # Icon drawn with the `image` option
         drawn_image_icon = qta.icon('ri.truck-fill',
--- a/README.md
+++ b/README.md
@@ -175,7 +175,8 @@ spin_icon = qta.icon('mdi.loading', color='red',
 spin_widget.setIcon(spin_icon)
 
 # Simple icon widget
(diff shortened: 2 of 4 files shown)


example.py
```diff
@@ -96,7 +96,7 @@ def __init__(self):
         lo.addWidget(iconwidget)
         lo.addWidget(QtWidgets.QLabel('IconWidget'))
         iconwidgetholder.setLayout(lo)
-        iconwidget2 = qta.IconWidget('mdi.web', color='blue')
+        iconwidget2 = qta.IconWidget('mdi.web', color='blue', size=QtCore.QSize(24, 24))
```
dalthviz, line 99: Since this changes the size, I think we will need to update the [examples screenshot/gif](https://github.com/spyder-ide/qtawesome/blob/master/qtawesome-screenshot.gif). Other option is to change this to pass the default value (16) so in that case there will be no need to update the screenshot.

Also, we need to update the corresponding text in the examples entry in the [`README.md`](https://github.com/spyder-ide/qtawesome/tree/master#examples) and the docs [`usage.rst`](https://github.com/spyder-ide/qtawesome/blob/master/docs/source/usage.rst) to follow this change

dalthviz (commented):
Thanks @StSav012 for your work here! Left a comment regarding some places and things that need an update to follow up with the change done at `example.py` and the usage there of the new `size` kwarg.

Other than that this LGTM :+1:

Thanks again!

example.py
StSav012: Thanks! I haven't thought of the two other files.

As a side note, why is `IconWidget` not mentioned in [the docs](https://qtawesome.readthedocs.io/en/latest/index.html)?

dalthviz: That's a good point @StSav012 ! Maybe we need to add some other params to the `automodule` directive or calling other directive in the [API Reference page](https://github.com/spyder-ide/qtawesome/blob/cd4768d3e7f5eb9a2ab7b639dfe866060a3dec63/docs/source/api_documentation.rst) so the `IconWidget` class docs appear? Let us know if you would like to help us checking that!

dalthviz (approved):
Thanks @StSav012 !

example.py
StSav012: I've tried to add `IconWidget` line before https://github.com/spyder-ide/qtawesome/blob/0bda7bf65c842b23711169d6563683b021300e13/qtawesome/__init__.py#L13 (or where's the best place for that?). The item appears in the docs. But `sphinx` is more clever than me, and it adds ugly empty lines about every single method of `QLabel`. As much as I'd like to tackle the problem, I have neither enough experience nor enough time these days. Maybe, closer to the weekend.

dalthviz: > or where's the best place for that?

Maybe we need to check if adding a `Custom widgets` section could be worthy but as a first approach makes sense to me :+1:

> As much as I'd like to tackle the problem, I have neither enough experience nor enough time these days. Maybe, closer to the weekend.

No worries, thank you for your time checking how the directive works and how we could improve the docs!

PR #241, opened 2023-07-11
PR: Update Codicons icons to 0.0.33

kumattau:
The bundled codicon-0.0.32 seems to have the following problem.

* md5sum does not match the official distribution.
* There is not enough charmap.

Therefore, it is recommended to update to the latest version 0.0.33.

The above problem is caused by the following in update_msc.

* The file version is not stable because the file is downloaded from master(main) branch.
* The [codicon.csv](https://github.com/microsoft/vscode-codicons/blob/main/dist/codicon.csv) is missing definitions and compatibility is broken because it is simply a list of glyph names. So, [the explicit version codicon.css](https://github.com/microsoft/vscode-codicons/blob/0.0.33/dist/codicon.css) need to be used instead.
* It seems that [packages.json of master(main) branch](https://github.com/microsoft/vscode-codicons/blob/main/package.json) is not following the release. So, it cannot be used for version check.


dalthviz:
Hi @kumattau sorry for the late response! I gave this a check and seems like now the [latest codicons release is 0.0.35](https://github.com/microsoft/vscode-codicons/releases/tag/0.0.35). Could it be possible for you to update this PR to that version? Let us know!

dalthviz:
Just in case, I created https://github.com/spyder-ide/qtawesome/pull/251 to update to the latest codicons version (0.0.35) so that PR will supersed this one. Also, I created a new issue to follow up your findings regardind the update command at https://github.com/spyder-ide/qtawesome/issues/252

dalthviz:
Merged via https://github.com/spyder-ide/qtawesome/pull/251

kumattau:
@dalthviz 
Sorry for late reply. Thank you for #251 !

Issue #165, opened 2021-07-12
Need a way to stop animation icon from calling _update


paljsingh:
Once a spin icon is added, there does not seem to be an easy way to stop it from calling icon update.
With the default interval of 10ms, it starts consuming a lot of cpu if there are more than a few such icons a page.

In my use case, I need spin icons temporarily to show items under progress, and later these are replaced with non-spin icons, however the spin icons do not get garbage collected and keep calling Spin._update every 10 ms.

Attaching a sample code to show the impact on cpu usage.
---
In idle case, the CPU usage is close to 0, and that remains close to 0 even after setting the icons to non spin ones.
However, once the button icons are set to spinner, CPU usage shoots up to 70% or above, and remains at the same level even if the buttons are later set to another icon.

```
import sys
from functools import partial

import qtawesome as qta
from PySide2.QtWidgets import QApplication, QWidget, QHBoxLayout, QPushButton, QVBoxLayout


class SpinIconTest(QWidget):

    def __init__(self):
        super().__init__()
        self.spin_buttons = list()
        self.setup_ui()

    def setup_ui(self):

        # layout..
        hbox = QHBoxLayout()
        vbox = QVBoxLayout()

        num_buttons = 10

        for i in range(num_buttons):
            button = QPushButton('button {}'.format(i), self)
            vbox.addWidget(button)
            self.spin_buttons.append(button)

        self.button1 = QPushButton('set spin icon!', self)
        self.button1.clicked.connect(partial(self.set_spin_icon))

        self.button2 = QPushButton('set to regular icon!', self)
        self.button2.clicked.connect(partial(self.set_regular_icon))

        hbox.addWidget(self.button1)
        hbox.addWidget(self.button2)

        container_widget = QWidget()
        container_widget.setLayout(vbox)
        hbox.addWidget(container_widget)
        self.setLayout(hbox)

        self.show()

    def set_spin_icon(self):
        for button in self.spin_buttons:
            button.setIcon(qta.icon("fa5s.spinner", animation=qta.Spin(button)))

    def set_regular_icon(self):
        for button in self.spin_buttons:
            button.setIcon(qta.icon("fa5s.check-circle"))


def main():

    app = QApplication([])
    ex = SpinIconTest()

    sys.exit(app.exec_())


if __name__ == '__main__':
    main()

```

paljsingh:
At Idle

[...]

PR #245, opened 2023-11-06
PR: Add API to start/stop animations to the Spin class

ccordoba12:
- This also adds a kwarg to decide if the animation should be automatically started (in some situations that's not desired).
- I realized this is necessary to stop the spinner based on a QtAwesome we added to Spyder in spyder-ide/spyder#21271. Without this, the spinner keeps running forever, even the widget that contains it is hidden.


dalthviz:
Should we mark this as something the fix for https://github.com/spyder-ide/qtawesome/issues/165 @ccordoba12 ?

ccordoba12:
Yep, you're totally right (I didn't know about that issue).

ccordoba12:
And my question for you: should we add documentation of this new feature? And if, where should I put it?

dalthviz:
> should we add documentation for this new feature?

That's a good point, I would say yes :+1:

> And if so, where should I put it?

Probably we could expand the `Animations` usage example at least:

https://github.com/spyder-ide/qtawesome/blob/d1371a7a3683805b9277059469dce98a04040504/README.md?plain=1#L158-L166

https://github.com/spyder-ide/qtawesome/blob/d1371a7a3683805b9277059469dce98a04040504/docs/source/usage.rst?plain=1#L171-L179

https://github.com/spyder-ide/qtawesome/blob/d1371a7a3683805b9277059469dce98a04040504/example.py#L123-L127

Maybe in the future we should try to expand the API reference to cover not only the `qtawesome/__init__.py` but also improve other places docstrings and use those to autogenerate more docs reference pages 🤔

ccordoba12:
> Probably we could expand the Animations usage example at least:

Done in the three places you referenced above.

dalthviz:
Note: Seems like on Windows with PyQt5 running the `example.py` script is failing with `exit code 127`. Other checks were failing due to the coverage upload step failing but a couple of reruns help with that 🤔

ccordoba12:
Fixed in my last commit by skipping running the example file when using PyQt5.

PR #250, opened 2023-12-07
[WIP] PR: Add codecov token as a repo secret

dalthviz:
Seems like [coverage upload CI step is failing](https://github.com/spyder-ide/qtawesome/actions/runs/7131431020/job/19420306201?pr=245) due to missing coverage token. Set token as secret and use it over the CI


dalthviz:
Closing since seems like with a CI rerun  made the coverage step to pass. Also, thinking about this, probably in the future a check to use a different service for code coverage could be worthy - codecov is kind of flaky sometimes 🤔

PR #263, opened 2024-06-18
Fix styling on PySide6

meowmeowahr:
# Issue
`QWidget.setStyle` on PySide6 only accepts a QStyle class, while PyQt5, PyQt6, and PySide2 can accept a QStyle or a string. This causes `qtawesome.dark()` / `qtawesome.light()` to fail on PySide6.

# Change
This change fixes the issue by using `QStyleFactory.create()` to set the application style in `styles.py`

### DIFF
--- a/qtawesome/styles.py
+++ b/qtawesome/styles.py
@@ -25,6 +25,7 @@ SOFTWARE.
 """
 
 from qtpy.QtGui import QPalette, QColor
+from qtpy.QtWidgets import QStyleFactory
 
 # Constant to reference default themes
 DEFAULT_DARK_PALETTE = 'Dark'
@@ -76,7 +77,7 @@ def dark(app):
     app.style().unpolish(app)
     app.setPalette(dark_palette)
 
-    app.setStyle('Fusion')
+    app.setStyle(QStyleFactory.create('Fusion'))
 
 
 def light(app):
@@ -124,4 +125,4 @@ def light(app):
     app.style().unpolish(app)
     app.setPalette(light_palette)
 
-    app.setStyle('Fusion')
+    app.setStyle(QStyleFactory.create('Fusion'))


dalthviz:
Hi @meowmeowahr thank you for submitting a fix! Do this issue raise with a specific version of PySide6? I tried to see if running things from master using PySide6 6.7.2 caused any issue related with the style but seems like I was unable to reproduce 🤔 Could you explain a little bit more how setting the app style was raising an issue for you or add some before/after screenshot preview to better understand how this change affects how things look? Thank you again for proposing changes and let us know!

meowmeowahr:
I can't seem to reproduce the issue anymore. I have tried all versions of PySide6 on both Linux and WIndows and it works fine with the code on master and my fork. 

Here is my test code
```python
from PySide6.QtCore import Qt, qVersion
from PySide6.QtWidgets import QWidget, QApplication, QPushButton, QVBoxLayout
import sys

import qtawesome

print("Qt", qVersion())

class TestWindow(QWidget):
    def __init__(self) -> None:
        super().__init__()
        
        self.vlayout = QVBoxLayout()
        self.setLayout(self.vlayout)

        self.light = QPushButton("Light")
        self.light.clicked.connect(lambda: qtawesome.light(app))
        self.vlayout.addWidget(self.light)

        self.dark = QPushButton("Dark")
        self.dark.clicked.connect(lambda: qtawesome.dark(app))
        self.vlayout.addWidget(self.dark)

        self.show()


if __name__ == "__main__":
    app = QApplication(sys.argv)
    win = TestWindow()
    app.exec()
```

dalthviz:
Oh thank you for the check! In that case it seems like the change is not needed, right? I would say we could close this one for now, but if you find a way to reproduce the issue let us know to be able to restart checking this work!

meowmeowahr:
Thanks, I don't think the change is needed as of now. I will close the PR.

PR #270, opened 2025-02-12
PR: Add fallback to try to load fonts from the system ones and add method to check info about bundled fonts

dalthviz:
Part of https://github.com/spyder-ide/spyder/issues/22499 following implementation idea suggested at https://github.com/spyder-ide/spyder/issues/22499#issuecomment-2415731253


qtawesome/__init__.py
```diff
@@ -81,6 +82,19 @@
 }
 
 
+def get_fonts_info():
+    """
+    Return tuple with information about the bundled fonts being used.
+
+    Tuple contains fonts directory path and a list with the fonts filenames.
```
ccordoba12, line 89:
```suggestion
    Notes
    -----
    The tuple contains the fonts directory path and a list of font filenames.
```

qtawesome/iconic_font.py
```diff
@@ -411,6 +411,13 @@ def hook(obj):
                 id_ = QFontDatabase.addApplicationFontFromData(data)
             font_data.close()
 
+            if id_ == -1 and os.name == "nt":
+                # Try to load font from system Fonts directory
+                windows_dir = os.environ.get("WINDIR")
+                if windows_dir:
```
ccordoba12, line 417: * Should we use `C:\Windows\Fonts` in case the `WINDIR` variable is not available?
* Also, did you manually test we can load the fonts from this directory now?

ccordoba12 (commented):
Just a small suggestion and a couple of questions for you @dalthviz.

qtawesome/iconic_font.py
dalthviz: > Should we use C:\Windows\Fonts in case the WINDIR variable is not available?

Well that could be another fallback indeed but it can end up also not working due to the Windows folder being in another drive. In any case, if you want to have a default I think we can then just remove the `if` and change line 416 to be something like:

```python
windows_dir = os.environ.get("WINDIR",  r"C:\Windows")
```
> Also, did you manually test we can load the fonts from this directory now?

Yep, I checked by hard-coding the `id_`  to be -1 instead of calling `addApplicationFontFromData` and then using `qta-browser`. That caused a `FontError` until I installed the fonts using the work over PR #271. Still, it would be nice to have a test for this. For the moment I'm unsure how to implement it but probably I would need some of the logic over PR #271 🤔

ccordoba12: > In any case, if you want to have a default I think we can then just remove the if and change line 416 to be something like:

No, you're right: `C:\Windows` could not exist either. I propose the following instead:

```python
if os.environ.get("WINDIR") is not None:
    windows_dir = os.environ.get("WINDIR")
elif os.path.isdir(r"C:\Windows"):
    windows_dir = r"C:\Windows"
else:
    windows_dir = ""

if windows_dir:
    ...
```

> probably I would need some of the logic over PR #271 🤔

Yeah, that's a good idea. You could use the script in #271 to install the fonts system-wide first, them make QtAwesome use that fonts with a special env var set by the test. So, it could be something like:

```python
if (id_ == -1 or os.environ.get("QTA_TEST_SYSTEM_FONTS")) and os.name == "nt":
    ...
```

dalthviz: End up implementing a tweaked version of your suggestions and adding a test that should only fail under windows for the moment. After, #271 gets merged I will update this one to make the test pass. Let me know what do you think

dalthviz:
@ccordoba12 I think this is ready for another review

qtawesome/tests/test_qtawesome.py
```diff
@@ -64,44 +64,78 @@ def test_bundled_font_user_installation():
...
 def test_bundled_font_system_installation():
     """
     Test that the bundled fonts can be installed on Windows for all users.
 
-    See spyder-ide/qtawesome#167 and spyder-ide/spyder#18642
+    See spyder-ide/qtawesome#244
+
+    Notes:
+        * When running this test is possible that a prompt for privileges
+        appears(UAC prompt).
```
ccordoba12, line 102:
```suggestion
    Notes
    -----
    * When running this test is possible that a prompt for privileges
      appears(UAC prompt).
```

Make this follow the Numpydoc format.

qtawesome/tests/test_qtawesome.py
```diff
@@ -64,44 +64,78 @@ def test_bundled_font_user_installation():
...
+        .stdout.replace("\n", "")
+        .replace(" ", "")
+    )
     for font_filename in fonts_expected:
         assert font_filename in fonts_result
 
 
+def test_font_load_from_system_fonts(monkeypatch):
+    """
+    Test that the bundled fonts can be accessed from the system fonts on Windows.
```
ccordoba12, line 125:
```suggestion
    Test that the bundled fonts can be accessed from the system fonts folder on
    Windows.
```

Clarify a bit more what the test does.

ccordoba12 (commented):
Some small docstring suggestions then this should be ready.

qtawesome/tests/test_qtawesome.py
```diff
@@ -64,44 +64,81 @@ def test_bundled_font_user_installation():
...
     """
     Test that the bundled fonts can be installed on Windows for all users.
 
-    See spyder-ide/qtawesome#167 and spyder-ide/spyder#18642
+    See spyder-ide/qtawesome#244
+
+    Notes
+    -----
+    * When running this test is possible that a prompt for privileges
+      appears(UAC prompt).
```
ccordoba12, line 103:
```suggestion
    * When running this test, it's possible that a prompt for privileges
      may appear (UAC prompt).
```

qtawesome/tests/test_qtawesome.py
```diff
@@ -64,44 +64,81 @@ def test_bundled_font_user_installation():
...
+    """
+    Test that the bundled fonts can be accessed from the system fonts folder on
+    Windows.
+
+    Notes
+    -----
+    * This test ensures that the logic to load fonts from the system fonts
+      only affects Windows even when it is being forced.
+    * When running this test is possible that a prompt for privileges
+      appears (UAC prompt).
```
ccordoba12, line 134:
```suggestion
    * When running this test, it's possible that a prompt for privileges may
      appear (UAC prompt).
```

ccordoba12 (commented):
Last suggestions (I didn't catch them in my previous review).

ccordoba12 (approved):

PR #271, opened 2025-02-13
PR: Add logic to install bundled fonts system wide on Windows

### LINKED ISSUE #244: Create script to allow users to install our fonts system-wide
After seeing the struggles users can go through to make our fonts work for their Windows user (see issue spyder-ide/spyder#21458) and taking into account that there won't be a fallback for Spyder 6, I think we need to provide another solution for them.

That's why I think we should add a script here (either in Python or a Batch script) that we can call from Spyder (or tell users to run in a terminal) to install our fonts as admin.

This SO post has some ideas to get started with this:

https://superuser.com/questions/201896/how-do-i-install-a-font-from-the-windows-command-prompt

dalthviz:
Fixes #244

### DIFF
--- a/qtawesome/__init__.py
+++ b/qtawesome/__init__.py
@@ -9,6 +9,7 @@ Font-Awesome and other iconic fonts for PyQt / PySide applications.
 .. autosummary::
    :toctree: _generate
 
+   install_bundled_fonts_system_wide
    icon
    load_font
    charmap
@@ -79,6 +80,10 @@ _MD5_HASHES = {
     "remixicon-2.5.0.ttf": "888e61f04316f10bddfff7bee10c6dd0",
     "codicon-0.0.35.ttf": "8478f5b3df2158f7e4864473e34efda1",
 }
+
+
+def install_bundled_fonts_system_wide():
+    _instance().install_fonts_system_wide()
 
 
 def has_valid_font_ids(inst):
--- a/qtawesome/iconic_font.py
+++ b/qtawesome/iconic_font.py
@@ -626,14 +626,14 @@ class IconicFont(QObject):
                 )
                 self._install_fonts(fonts_directory, system_wide=True)
             else:
-                # Call `qta-install-fonts-system-wide` using `runas` to prompt user for admin elevation
+                # Call `qta-install-fonts-all-users` using `runas` to prompt user for admin elevation
                 warnings.warn(
                     "Admin privileges are needed to install the bundled fonts. "
                     "A prompt to get them will be shown and after that this "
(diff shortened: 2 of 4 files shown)


qtawesome/__init__.py
```diff
@@ -81,6 +82,10 @@
 }
 
 
+def install_bundled_fonts_system_wide():
```
ccordoba12, line 85: Here you can find some ideas on how to make this require elevated privileges automatically (in case the user is not an admin):

https://stackoverflow.com/questions/19672352/how-to-run-script-with-elevated-privilege-on-windows

qtawesome/iconic_font.py
```diff
@@ -616,6 +616,24 @@ def set_custom_icon(self, name, painter):
...
+                fonts_directory = os.path.join(
+                    os.path.dirname(os.path.realpath(__file__)), "fonts"
+                )
+                self._install_fonts(fonts_directory, system_wide=True)
+            else:
+                # Call `qta-fonts` using `runas` to prompt user for admin elevation
+                warnings.warn(
+                    "Need for admin privileges to install bundled fonts. "
+                    "A prompt to get admin privileges and relaunch the command "
+                    "will be shown..."
```
ccordoba12, line 633:
```suggestion
                    "Admin privileges are needed to install the bundled fonts. "
                    "A prompt to get them will be shown and after that this "
                    "command will be relaunched..."
```

Improve wording for this message.

setup.py
```diff
@@ -51,6 +51,9 @@
         "update_msc": setupbase.UpdateCodiconCommand,
     },
     entry_points={
-        "console_scripts": ["qta-browser=qtawesome.icon_browser:run"],
+        "console_scripts": [
+            "qta-browser=qtawesome.icon_browser:run",
+            "qta-fonts=qtawesome:install_bundled_fonts_system_wide",
```
ccordoba12, line 56: I think this script should be named `qta-install-fonts-system-wide` (or something similar) to be more specific about what it does.

I know the name is quite long (so, I'm open to suggestions), but the command not only installs our fonts, it does it for the system, not for the current user.

dalthviz: Maybe something like `qta-install-fonts-allusers` or `qta-install-fonts-system`? For the moment, will change the name to the one you suggested but let me know if any other alternative is better

ccordoba12: > Maybe something like qta-install-fonts-allusers

This sounds good to me, but adding a dash between `all` and `users`:

    qta-install-fonts-all-users

ccordoba12 (approved):

PR #273, opened 2025-02-19
PR: Return original font directory when font user installation load fails

### LINKED ISSUE #264: FileNotFoundError on fonts due to Windows Font Blocking policy
I have a PySide6 application using QtAwesome which has worked fine on every machine I have installed it on up until now. On this new machine, the application fails to start with a FileNotFoundError on fontawesome4-webfont-4.7.ttf. I have created a small test application, which exhibits the same problem...

```
import sys
import qtawesome as qta
from PySide6.QtWidgets import (QApplication, QWidget, QGridLayout,
                               QToolButton, QLabel)

app = QApplication(sys.argv)

window = QWidget()
layout = QGridLayout(window)
names = ['fa5s.grip-horizontal', 'fa5s.file-export', 'fa5s.file-import']
for i, name in enumerate(names):
    qicon = qta.icon(name, scale_factor=1.3)
    btn = QToolButton()
    btn.setIcon(qicon)
    btn.setIconSize(QSize(32, 32))
    layout.addWidget(btn, i, 0)
    layout.addWidget(QLabel(name), i, 1)
window.show()

sys.exit(app.exec_())
```

The error I get is this...
```
Traceback (most recent call last):
  File "C:\...\pydev\qta_test\qta_test.py", line 12, in <module>
    qicon = qta.icon(name, scale_factor=1.3)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\...\pydev\qta_test\.venv\Lib\site-packages\qtawesome\__init__.py", line 234, in icon
    return _instance().icon(*names, **kwargs)
           ^^^^^^^^^^^
  File "C:\...\pydev\qta_test\.venv\Lib\site-packages\qtawesome\__init__.py", line 130, in _instance
    _resource['iconic'] = IconicFont(*_BUNDLED_FONTS)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\...\pydev\qta_test\.venv\Lib\site-packages\qtawesome\iconic_font.py", line 337, in __init__
    self.load_font(*fargs)
  File "C:\...\pydev\qta_test\.venv\Lib\site-packages\qtawesome\iconic_font.py", line 377, in load_font
    with open(os.path.join(directory, ttf_filename), 'rb') as font_data:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\...\\AppData\\Local\\Microsoft\\Windows\\Fonts\\fontawesome4-webfont-4.7.ttf'
```
I'm using a venv, which works on every other machine I've installed on. The fontawesome4-webfont-4.7.ttf file is not present on any of the machines in the AppData path in the traceback. It IS in .venv\Lib\site-packages\qtawesome\fonts on every machine, including the one on which the error is occurring.

This is on Windows 11 and Python 3.12.5 installed from the python.org installer.

dalthviz:
Fixes #264


qtawesome/iconic_font.py
```diff
@@ -725,7 +725,17 @@ def _install_fonts(self, fonts_directory, system_wide=False):
                             # Possible permission error when trying to remove
                             # a font which potentially is already in use
                             # See spyder-ide/qtawesome#236
-                            continue
+                            pass
+                        # Regardless of the font file being actually removed
+                        # since the load failed we need to return here where
+                        # the font file is originally located. With that a
+                        # `FontError` can be eventually raised with useful
```
ccordoba12, line 732: Instead of returning `fonts_directory` below, why don't we raise a `FontError` there, saying that it was not possible to load a specific font? The error message could also use `get_fonts_info` to tell users what fonts they need to install manually and where they are located.

dalthviz: After giving it some thought, if we raise `FontError` here the fallback logic to use system wide fonts will never be reached (since then always an error here will be raised). Maybe we should show instead a warning here and only in case we are unable to detect the font that failed being loaded is available from the system fonts directory? What do you think @ccordoba12 ?

dalthviz: Last commit implements the idea above but let me know what you think

ccordoba12:
Sorry for closing this PR (I don't know why it happened).

qtawesome/iconic_font.py
ccordoba12: Ok, that looks better to me. But then I think we don't need to return `fonts_directory` below, but a simple `return`. And the comment you added here should be reworded, right?

qtawesome/iconic_font.py
```diff
@@ -725,7 +725,27 @@ def _install_fonts(self, fonts_directory, system_wide=False):
...
+                        # information if necessary (i.e not being able to load
+                        # bundled fonts pointing that the fonts are located on
+                        # the QtAwesome `fonts` directory and manual
+                        # configuration is needed to allow them). Also, if we
+                        # don't detected the font being available from the
+                        # system fonts directory a warning with a more specific
+                        # suggestion (install the font manually) is shown.
```
ccordoba12, line 739:
```suggestion
                        # Regardless whether the font file is actually removed,
                        # we need to return here where the font file is
                        # originally located since the load failed. With that,
                        # a `FontError` can be eventually raised with useful
                        # information (i.e not being able to load bundled
                        # fonts, saying that the fonts are located in the
                        # QtAwesome `fonts` directory and manual configuration
                        # is needed to allow them). Also, if we don't detected
                        # that the font being available from the system fonts
                        # directory, a warning with a more specific suggestion
                        # (i.e. to install the font manually) is shown.
```

qtawesome/iconic_font.py
```diff
@@ -725,7 +725,27 @@ def _install_fonts(self, fonts_directory, system_wide=False):
...
+                        # See spyder-ide/qtawesome#264
+                        if not os.path.isfile(os.path.join(system_fonts_dir, filename)):
+                            warnings.warn(
+                                "Font at '{0}' could not be loaded. If possible, "
+                                "install it system wide to make it available. "
+                                "Font source file to be manually installed can "
+                                "be found at '{1}'".format(dst_path, src_path)
```
ccordoba12, line 746:
```suggestion
                                "Font at '{0}' could not be loaded. If "
                                "possible, install it system wide to make it "
                                "available to QtAwesome. The source file to "
                                "be manually installed can be found at "
                                "'{1}'".format(dst_path, src_path)
```

ccordoba12 (commented):
Some improvements to the comment and warning text, then this should be ready.

ccordoba12 (approved):

PR #276, opened 2025-02-21
PR: Update Codicon to 0.0.36

### LINKED ISSUE #275: Update Codicon to 0.0.36 and fix update script
With version 0.0.36 the codicon repo removed the `dist` directory that we were using to update the font. Now the `.ttf` file needs to be downloaded from the release assets (for example https://github.com/microsoft/vscode-codicons/releases/download/0.0.36/codicon.ttf) and the charmap needs to be build probably using the [`mapping.json` file](https://github.com/microsoft/vscode-codicons/blob/main/src/template/mapping.json)

dalthviz:
Fixes #275


ccordoba12 (approved):
Thanks @dalthviz!

PR #283, opened 2025-03-28
[WIP] PR copy fonts in WSL

jsbautista:


dalthviz:
@jsbautista is this patch still required? Is something missing for it to work or maybe this approach ended up not being the correct one? Just wondering if this should be included in the next bug fix release (i.e 1.4.1), or maybe a minor one (e.g 1.5.0) or maybe it just should be closed. Let me know!

dalthviz:
Checking the Spyder issue tracker I think this could be related with https://github.com/spyder-ide/spyder/issues/23840 ?

Since the actual reason for the failure was the theme selection being set to the Spyder old/fallback theme seems to me that we could close this, right @ccordoba12 ?

ccordoba12:
Yes, @jsbautista checked that this is not necessary for WSL so it can be closed (and the Spyder issue too).

PR #288, opened 2025-08-01
[WIP] PR: Add PySide6 to CI

dalthviz:


.github/workflows/linux-tests.yml
```diff
@@ -23,6 +23,8 @@ jobs:
       matrix:
         PYTHON_VERSION: ['3.7', '3.10']
```
ccordoba12, line 24: Could you open a new issue to remove support for Python 3.7 and 3.8? We don't need to support them anymore because they reached end-on-life but we can't make that change in bugfix release (i.e. 1.4.1).

PR #298, opened 2025-12-17
PR: Fix icon center when using a custom rect definition

fdklgbh:
Fix the incorrect calculation of the icon's center coordinates (x_center, y_center) when QRect's left or top is not zero, which causes animation anomalies.

### DIFF
--- a/qtawesome/animation.py
+++ b/qtawesome/animation.py
@@ -30,8 +30,8 @@ class Spin:
                 timer.start(self.interval)
         else:
             timer, angle, self.step = self.info[self.parent_widget]
-            x_center = rect.width() * 0.5
-            y_center = rect.height() * 0.5
+            x_center = rect.x() + rect.width() * 0.5
+            y_center = rect.y() + rect.height() * 0.5
             painter.translate(x_center, y_center)
             painter.rotate(angle)
             painter.translate(-x_center, -y_center)


dalthviz:
Hi @fdklgbh thank you for your work here and sorry for the late response! Could it be possible for you to give us a code example to better understand the issue this fixes please?

Note: Closing and reopening the PR to retrigger CI

dalthviz:
Note: Updating with latest master should fix the failing macOS CI here

fdklgbh:
Hi, @dalthviz, This is an example that should provide a better understanding of the issue
```python
import sys

from qtpy import QtWidgets, QtCore, QtGui

import qtawesome as qta
from qtpy.QtCore import QTimer, QRect


class RectIconWidget(qta.IconWidget):
    def __init__(self, *args, draw_rect=None, **kwargs):
        super().__init__(*args, **kwargs)
        self._draw_rect = draw_rect or QtCore.QRect()

    def paintEvent(self, event):
        if not self._draw_rect.isValid():
            return super().paintEvent(event)
        else:
            painter = QtGui.QPainter(self)
            self._icon.paint(painter, self._draw_rect, QtCore.Qt.AlignCenter)


class FixedSpin(qta.Spin):
    def setup(self, icon_painter, painter, rect):
        if self.parent_widget not in self.info:
            super().setup(icon_painter, painter, rect)
        else:
            timer, angle, self.step = self.info[self.parent_widget]
            x_center = rect.x() + rect.width() * 0.5
            y_center = rect.y() + rect.height() * 0.5
            painter.translate(x_center, y_center)
            painter.rotate(angle)
            painter.translate(-x_center, -y_center)


class Demo(QtWidgets.QWidget):
    def __init__(self):
        super().__init__()
        layout = QtWidgets.QHBoxLayout(self)
        self.setFixedSize(400, 300)
        spin_widget_1 = RectIconWidget(draw_rect=QRect(10, 10, 20, 20))
        animation = qta.Spin(spin_widget_1)
        icon_1 = qta.icon('mdi.loading', color='black', animation=animation)
        spin_widget_1.setIcon(icon_1)
        spin_widget_1.setStyleSheet("background-color: red;")
        spin_widget_1.setFixedSize(50, 50)

        spin_widget_2 = RectIconWidget(draw_rect=QRect(10, 10, 20, 20))
        fix_animation = FixedSpin(spin_widget_2)
        icon_2 = qta.icon('mdi.loading', color='black', animation=fix_animation)
        spin_widget_2.setIcon(icon_2)
        spin_widget_2.setStyleSheet("background-color: yellow;")
        spin_widget_2.setFixedSize(50, 50)

        layout.addWidget(spin_widget_1)
        layout.addWidget(spin_widget_2)
        QTimer.singleShot(5000, animation.stop)
        QTimer.singleShot(5000, fix_animation.stop)


if __name__ == '__main__':
    app = QtWidgets.QApplication(sys.argv)
    window = Demo()
    window.show()
    sys.exit(app.exec())
```

qtawesome/animation.py
```diff
@@ -30,8 +30,8 @@ def setup(self, icon_painter, painter, rect):
                 timer.start(self.interval)
         else:
             timer, angle, self.step = self.info[self.parent_widget]
-            x_center = rect.width() * 0.5
-            y_center = rect.height() * 0.5
+            x_center = rect.x() + rect.width() * 0.5
```
dalthviz, line 33: Please add here a comment pointing to the example code comment provided. Something like:

```suggestion
            # `x` and `y` addition needed to allow proper center computation with
            # custom icon rect definitions.
            # See https://github.com/spyder-ide/qtawesome/pull/298#issuecomment-3793713831 
            x_center = rect.x() + rect.width() * 0.5
```

dalthviz (commented):
Thank you @fdklgbh for the code example! Although I think the changes here are only needed then in case you use a custom `IconWidget` class definition, seems to me that it shouldn't be a problem to add `x()` and `y()` in general (which should be equivalent to adding `0` to the computation of the icon center when using `IconWidget`)

Left a comment with a suggestion to add an inline comment referencing the usage of `x()` and `y()` rect values. Also, this needs to be on top of latest master (to make the macOS CI pass).

As a side note, thinking about the example provided, another change we could do here is to include the `draw_rect` kwarg in the `IconWidget` definition so you wouldn't need to subclass `IconWidget` 🤔However. just to be sure that could be worthy, what do you think about that @ccordoba12 ?

qtawesome/animation.py
fdklgbh: Hi @dalthviz, The `IconWidget` shown here is just an example—in practice, it could be any other component. The `draw_rect` parameter is included solely for demonstration; in real usage, such layout details are typically handled internally by the component itself.

dalthviz: Maybe I'm not understanding this correctly then 🤔 Doesn't the need for `x` and `y` only happens if someone passes here a custom rect? Maybe do you have an example of a case where a rect with non-zero `x`/`y` values ends up being passed without needing to subclass QtAwesome classes? If the phrasing of the comment suggestion above doesn't describe correctly the situation here, how the the following sounds to you?:

```python
  # `x` and `y` addition needed to allow proper center computation with
  # custom rect definitions (with non-zero `x`/`y` values).
  # See https://github.com/spyder-ide/qtawesome/pull/298
```

Also, could you describe the way you use QtAwesome, please? Further thinking on this seems to me that current QtAwesome logic will never get something different from 0 from `x` and `y` here, right? :thinking:

fdklgbh: This is a demo with not `draw_rect`
```python
import sys
from typing import Union

from PySide6.QtCore import QRectF
from PySide6.QtGui import QIcon, QPainter, Qt
from PySide6.QtWidgets import QPushButton, QApplication

from qtpy import QtWidgets
import qtawesome as qta
from qtpy.QtCore import QTimer


class PushButton(QPushButton):

    def icon(self):
        return self._icon

    def setIcon(self, icon: Union[QIcon, str]):
        self.setStyle(QApplication.style())
        self._icon = icon or QIcon()
        self.update()

    def paintEvent(self, e):
        super().paintEvent(e)
        if self.icon().isNull():
            return

        painter = QPainter(self)
        painter.setRenderHints(QPainter.Antialiasing |
                               QPainter.SmoothPixmapTransform)

        w, h = self.iconSize().width(), self.iconSize().height()
        # Some custom logic returns x and y
        x, y = 15, 15
        self.icon().paint(painter, QRectF(x, y, w, h).toRect(), Qt.AlignCenter)


class FixedSpin(qta.Spin):
    def setup(self, icon_painter, painter, rect):
        if self.parent_widget not in self.info:
            super().setup(icon_painter, painter, rect)
        else:
            timer, angle, self.step = self.info[self.parent_widget]
            x_center = rect.x() + rect.width() * 0.5
            y_center = rect.y() + rect.height() * 0.5
            painter.translate(x_center, y_center)
            painter.rotate(angle)
            painter.translate(-x_center, -y_center)


class Demo(QtWidgets.QWidget):
    def __init__(self):
        super().__init__()
        layout = QtWidgets.QHBoxLayout(self)
        self.setFixedSize(400, 300)
        spin_widget_1 = PushButton(text='btn1', parent=self)

        animation = qta.Spin(spin_widget_1)
        icon_1 = qta.icon('mdi.loading', color='black', animation=animation)
        spin_widget_1.setIcon(icon_1)
        spin_widget_1.setFixedSize(100, 100)

        spin_widget_2 = PushButton(text='btn2', parent=self)
        fix_animation = FixedSpin(spin_widget_2)
        icon_2 = qta.icon('mdi.loading', color='black', animation=fix_animation)
        spin_widget_2.setIcon(icon_2)
        spin_widget_2.setFixedSize(100, 100)

        layout.addWidget(spin_widget_1)
        layout.addWidget(spin_widget_2)
        QTimer.singleShot(5000, animation.stop)
        QTimer.singleShot(5000, fix_animation.stop)


if __name__ == '__main__':
    app = QtWidgets.QApplication(sys.argv)
    window = Demo()
    window.show()
    sys.exit(app.exec())
```

dalthviz: Thank you for the new example @fdklgbh ! If the comment suggested above still doesn't makes sense to you that's fine. The only pending thing here then is to update this PR with master so CI checks on macOS pass, so will do that and as soon CI is green will merge :+1:

dalthviz (approved):
Thank you @fdklgbh !

Issue #310, opened 2026-07-16
QtMDi (third-party Material Symbols extension) reached v2.0.x — Simple Icons, lazy-loading, font-weight fixes


o-murphy:
Hi maintainers,

I maintain [QtMDi](https://github.com/o-murphy/QtMDi), an extension for QtAwesome that adds Google Material Symbols (`mds`/`mdf` prefixes) as dynamically-loaded fonts via `qtawesome.load_font`. It's been around since 2023 and just reached **v2.0.x**, but as far as I can tell it isn't referenced anywhere in QtAwesome's README, docs, or wiki - so I wanted to introduce it here in case it's worth a mention for users looking to extend QtAwesome with Material Symbols.

**What QtMDi adds:**

- **`mds`/`mdf` prefixes** - Google's official Material Symbols, in both the static "symbols" style and the variable-font style, across all weights.
- **`si` prefix** (new in v2.0.x) - [Simple Icons](https://github.com/simple-icons/simple-icons-font) brand/logo icons, e.g. `qtawesome.icon("si.github")`.
- **Lazy loading** (new in v2.0.x) - `qtmdi.load(app)` is lazy by default: it only registers which fonts *could* be loaded, and defers actually reading the font file until an icon with that prefix is first requested. Keeps memory/startup cost low for apps that only use a handful of styles/weights.
- **Fixed `font-weight` handling for `mds` icons** (v2.0.x) - variable-width Material Symbols weights (e.g. `mds-rounded-700.home`) previously had a weight-selection bug in Qt; now fixed at the font level (family-name patching during font fetch/build).

**How QtMDi stays up to date:** the fonts aren't manually bumped - there's a scheduled GitHub Actions workflow (`symbols-update`) that fetches the Material Symbols variable fonts straight from Google's [material-design-icons](https://github.com/google/material-design-icons) repo and the Simple Icons font from its [GitHub Releases](https://github.com/simple-icons/simple-icons-font/releases), regenerates per-weight static fonts and the icon charmap, and writes a manifest checksum of the font files. If (and only if) the fonts actually changed, the workflow cuts a new release, which triggers automatic publishing to PyPI. So QtMDi tracks upstream icon sources continuously with no manual intervention needed for routine icon updates - only code/API changes (like the ones above) require a manual release.

[...]

PR #312, opened 2026-07-17
PR: add support for 3rd-party font providers

o-murphy:
## Summary

Adds a plugin system that lets third-party packages register extra iconic fonts with qtawesome, without needing changes on the qtawesome side.

- `qtawesome/__init__.py`: `_get_external_fonts()` discovers packages that expose an entry point in the `qtawesome.font_providers` group (via `importlib.metadata.entry_points`), calls each one, and loads the returned fonts alongside the bundled ones. Falls back to the pre-3.10 `entry_points()` API for Python 3.9 compatibility.
- `font_provider_template/`: reference implementation showing how to build a provider package - entry point registration in `pyproject.toml`, a `get_fonts()` callable returning `(prefix, ttf_filename, charmap_filename, directory)` tuples, and a working example bundling the Simple Icons font (`si` prefix) under its own CC0 1.0 license, separate from the template's MIT license.

**Note:** this is a proposal for the concept. The project's main documentation has not been updated yet - it will only be updated once the approach is agreed upon, to avoid documenting an API that might still change during review.

`font_provider_template/` currently lives inside this repo for review convenience; once the concept is agreed upon, it could be moved out to its own repository (or a dedicated branch) so that anyone wanting to build a font provider can start from it directly, without cloning qtawesome itself.

## Test plan

```bash
git clone https://github.com/o-murphy/qtawesome -b feat/font_providers
cd qtawesome
python3 -m venv .venv
. .venv/bin/activate
pip install -e .
pip install -e font_provider_template
qta-browser  # the 'si' prefix is now available
```

Relative to: #310


dalthviz:
Hi @o-murphy thank you for submitting an implementation idea to better support custom fonts loading! Thinking about this, I don't think loading external fonts automatically is a good idea to be honest so I would say the actual load of the external fonts should be done manually. I would suggest making the `_get_external_fonts` public (i.e `get_external_fonts`) and create an example that uses an external font to manually load it, so the user has to actually do something like:

```python
external_fonts = qta.get_external_fonts()

for external_font in external_fonts:
    self.load_font(*external_font)
```

Also, I think it could be worthy to create some utility function that helps users validate font files corruption in case someone wants to check that in a similar way as we do for bundle fonts via a hash.

Also, what do you think about this @ccordoba12 ?

o-murphy:
> Hi @o-murphy thank you for submitting an implementation idea to better support custom fonts loading! Thinking about this, I don't think loading external fonts automatically is a good idea to be honest so I would say the actual load of the external fonts should be done manually. I would suggest making the `_get_external_fonts` public (i.e `get_external_fonts`) and create an example that uses an external font to manually load it, so the user has to actually do something like:
> 
> 
> 
> ```python
> 
> external_fonts = qta.get_external_fonts()
> 
> 
> 
> for external_font in external_fonts:
> 
>     self.load_font(*external_font)
> 
> ```
> 
> 
> 
> Also, I think it could be worthy to create some utility function that helps users validate font files corruption in case someone wants to check that in a similar way as we do for bundle fonts via a hash.
> 
> 
> 
> Also, what do you think about this @ccordoba12 ?

I’ve already proposed a good solution with lazy-load by prefix + entry-point on issue #313  so public `get_external_font` is not really needed.

dalthviz:
> I’ve already proposed a good solution with lazy-load by prefix + entry-point on issue https://github.com/spyder-ide/qtawesome/issues/313 so public get_external_font is not really needed

So I guess we should focus then on that approach there first. After that, we could then revisit this and see what needs changes here and how things then get loaded just when the user requests it. right?

o-murphy:
> > I’ve already proposed a good solution with lazy-load by prefix + entry-point on issue #313 so public get_external_font is not really needed
> 
> So I guess we should focus then on that approach there first. After that, we could then revisit this and see what needs changes here and how things then get loaded just when the user requests it. right?

Exactly

ccordoba12:
> Also, what do you think about this @ccordoba12 ?

Looks good to me, thanks @o-murphy for working on this!

